PR #63, opened 2016-12-19
pgversion

pvacca:
PG_VERSION var for 9.5 container, and re-order yum install to prep for postgis split.


jmccormick2001:
merged

PR #73, opened 2017-01-10
Gh48 kubernetes tests

pvacca:
Unit tests for Kubernetes: basic, master-nfs, master-replica, backup, restore


jmccormick2001:
please rebase and add the openshift nfs examples to this PR when you get a chance...I like the idea of makiing the NFS path and server more configurable like this but want all the examples to match.

jmccormick2001:
reworked NFS path and Pv and PVC creation making this PR no longer valid.

PR #76, opened 2017-01-13
bin/pgbouncer/pgbouncer-watch.sh: call `kubectl`, not `oc` inside kube_failover()

dkoston:
Fixes a bug where `oc` was called instead of `kubectl` when trying to failover using kubernetes.

also removes unnecessary spaces at end of lines.


jmccormick2001:
thanks for fixing this!

PR #86, opened 2017-02-06
stricter start.sh to ensure all vars are set

pvacca:
Added stricter set statements to start.sh - fail on error or undefined variable. 
Resolved setenv.sh problem where chown is called before directory is created.
Rebased against 1.2.8 master.


pvacca:
Rebase has made this impossible to see actual change.

PR #87, opened 2017-02-10
Stricter start.sh to ensure all vars are set

pvacca:
Rebased against 1.2.8 master.
Added stricter set statements to start.sh - fail on error or undefined variable.
Resolved setenv.sh problem where chown is called before directory is created.


jmccormick2001:
out of date at this point.

PR #95, opened 2017-03-08
GH49 additional golang tests

pvacca:
Hi, Jeff, I had a little downtime, so I added a test for pgbouncer, and cleaned up some other areas of the code.


jmccormick2001:
thanks Phil, I'll merge this in next week when I get back from PTO.

jmccormick2001:
I noticed that you made changes to the Dockerfile and start script, I wanted to make sure you tested the pgbouncer examples for openshift and/or kube after making these changes?  if not, please test to make sure they still work after this change.  thanks.

jmccormick2001:
you might want to split out the Dockerfile change from the test framework changes to make the PR easier to get merged....for Dockerfile changes it means all dockerfiles for all versions would need to be changed and then tested across the various versions.

pvacca:
Ok, I will split this. I had made the changes to the pgbouncer container because I cannot currently run the example with plain docker. It looks like the problem is the new subfolder - bouncerconfig - expected under pgconf. The directory does not exist in the image, and doesn't get mounted in the docker the example.

jmccormick2001:
when you come back to the pgbouncer stuff in a new PR, take a look at the backrest example, in that example it creates a subdir in the volume prior to the container running, something like that will need to be added to the pgbouncer example.

jmccormick2001:
this is out of date at this point and would need to be redone.

PR #104, opened 2017-03-31
Add option for custom WAL location

the1forte:
Hi Jeff,

Created a new var called XLOGDIR to use for a custom WAL location for initdb. I threw in a docker example as a bonus  :)


jmccormick2001:
looks good Chris, thanks!...merged.  this missed the 1.3.0 release we cut yesterday, let me know if you require it sooner than next release.

PR #106, opened 2017-04-06
Backrest restore

the1forte:
Hi Jeff,

Here's a brand new container to try out. It's only purpose is to mount /pgdata, /pgconf and /backrestrepo so you can perform a pgbackrest restore command. I've included a K8s example. Let me know if you want to run through this with me later.

thanks,
Chris


jmccormick2001:
let me know when you have this one ready to merge

the1forte:
Jeff,

New example is ready! It runs as a job, this is the best fit for our environment. I even included the DELTA option. Hope you get a chance to try this out soon.

PR #113, opened 2017-04-11
Add line explaining necessary restart

xenophenes:


jmccormick2001:
for the .bashrc changes to take effect, you can just log out and log back in without restarting?

xenophenes:
Ah - I didn't try that - not much different from restarting but faster I suppose

jmccormick2001:
thanks Sarah!

PR #121, opened 2017-04-14
Misc. changes

xenophenes:
Remove duplicate section, fix lists, remove unnecessary "link:" occurences


xenophenes:
"for completness sake" isn't a proper sentence, changed to "for thoroughness" 
Removed all instances of "link:" where it's not redirecting to an internal file or displaying [visible text to hide an URL], as if neither of those mentioned cases are true then it is translated literally as:
link:insertlinkhere.com
Corrected any typos or misspellings as I came across them.

jmccormick2001:
thanks!

PR #135, opened 2017-05-02
Misc. grammatical fixes & correct accuracy of all docker examples

xenophenes:


xenophenes:
One aspect I'm confused on - 

master-replica - Running PostgreSQL in Master-Slave Configuration

and

master-replica - Creating a master and replica database cluster

Both refer to the container master-replica. Is that supposed to be the case?

xenophenes:
Also - 

Start the crunchy-dns server by running its container as follows:
....
cd $CCPROOT/examples/docker/dns
./run.sh
....

This folder does not exist. Is this deprecated?

xenophenes:
On trying to run the "watch" section - 

docker: Error response from daemon: manifest for crunchydata/crunchy-watch:centos7-9.5-1.2.8 not found.

Unsure how to solve this.

jmccormick2001:
yep, crunchy-dns is deprecated, so you can remove any references to it that
still exist.

jmccormick2001:
that is a reference to an older image tag, it should point to
centos7-9.5-1.4.0

jmccormick2001:
the first description of the master-replica example I think is the best,
that example creates/starts a postgres master container and also a replica
container, they essentially form a postgres 'cluster', the replica attempts
to contact the master to begin replication of state from the master to the
replica.

PR #138, opened 2017-05-04
Merge templates & origin with openshift docs

xenophenes:


xenophenes:
When comparing the documents, most of the tips provided in origin.asciidoc were the same as for openshift so I only added those that weren't present.

PR #146, opened 2017-05-09
help.md docs review

xenophenes:


xenophenes:
Is it preferred to leave "Usage:" sections summarized at "see Crunchy docs"? Or would it be better to explain examples in that section for each container?

jmccormick2001:
I'd probably go with 'see Crunchy docs' for now

xenophenes:
OK. All containers listed in https://github.com/CrunchyData/crunchy-containers/tree/master/rhel7/9.6 and /rhel7/ have an associated help file under /docs/. What sections of the help files would be best to add more information to?

xenophenes:
Also, RHEL Server version 7.3 is out. Should all the version numbers be replaced with 7.2 (in both the Dockerfiles and help files)?

jmccormick2001:
for now we'll stick with using and referencing RHEL 7.2, at some point in the future we'll take the plunge and upgrade.

jmccormick2001:
for the help file content, I'd vote for just the minimal amount of docs there, if something can be copied from the containers.asciidoc then that would be fine, if content doesn't exist, then I'd just leave it blank for now or 'refer to the crunchy docs' as a default placeholder.

xenophenes:
OK - please let me know if there's anything else you need in reference to the help files (in terms of editing/adding). 

Verified that all files exist and are correctly referenced with accurate information.

PR #154, opened 2017-05-20
misc minor fixes & add container info for crunchy-pgadmin4, crunchy-b…

xenophenes:
…ackup

### DIFF
--- a/docs/build-docs.sh
+++ b/docs/build-docs.sh
@@ -42,6 +42,17 @@ asciidoc \
 -a toc-placement=right \
 ./install.asciidoc
 
+asciidoctor-pdf ./dedicated.asciidoc --out-file ./pdf/dedicated.pdf
+
+asciidoc \
+-b bootstrap \
+-f ./demo.conf \
+-o ./htmldoc/dedicated.html \
+-a toc2 \
+-a toc-placement=right \
+./dedicated.asciidoc
+
+
 asciidoctor-pdf ./install.asciidoc --out-file ./pdf/install.pdf
 
 asciidoc \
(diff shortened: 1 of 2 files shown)


xenophenes:
crunchy-backup & crunchy-pgadmin4 were the only two with exact matches to containers in containers.asciidoc. 

- Please let me know if the file changes I made were the result you were looking for.
- crunchy-master, crunchy-restore and crunchy-replica I wanted to confirm their matches:
       crunchy-master & crunchy-replica looks like they're referring to master-replica found in docker.asciidoc
       crunchy-restore looks like it is referring to master-restore found in docker.asciidoc

xenophenes:
Regarding the second commit "tentative beginnings":

- Do you have to define the code that appears in dedicated.pdf similarly to containers.pdf, docker.pdf, etc? Or is that code automatically generated?
- In build-docs.sh, it references html files within the htmldoc directory but those don't seem to be present; is it simply stating that *would* be the name of the generated html document had you generated that document in html form?
- How do you define a specific stylesheet? I'm assuming within the block referencing dedicated.asciidoc within build-docs.sh, but it seems like the only "stylesheet" referenced within that sh file is -b bootstrap which looks like it might be referencing the multitude of asciidoctor-bootstrap.css files? But how is -b bootstrap mapped to all 15 or so bootstrap css files?

Note: I'd still like to know about these just to better understand the existing document builder tools, however; Paul didn't realize there was already a set of existing generated PDFs and he said the style that exists now will work.

xenophenes:
Please check out the complete document & make sure that it's correct. I believe everything is complete - I just need to review the examples themselves and ensure they're working properly.

jmccormick2001:
looking good, there is a section that describes the crunchy-postgres container because it is used in various templates, thats good...however, I don't see a similar section describing the crunchy-pgadmin4 or crunchy-backup containers, these are used in some of the templates as well, to match the outline of the document you would want to add sections for those other containers as well prior to describing the templates.

xenophenes:
Done!

PR #167, opened 2017-06-15
permission issues with Kubernetes and HEKETI

FaKod:
Hi,

I had a permission issue with Kubernetes and a mounted GlusterFS/HEKETI volume.
That's because you are doing chown in Dockerfile and executing setup.sh as user 26

To allow cluster filesystems to do proper user id mapping, chown should be executed in a root script.

The changes here works for me (however, I'm just testing), but may not be "elegant" enough...

### DIFF
--- a/bin/postgres/pathinit.sh
+++ b/bin/postgres/pathinit.sh
@@ -1,0 +1,8 @@
+#!/bin/bash
+
+chown -R postgres /pgdata
+chmod 700 /pgdata
+
+chown -R postgres /pgwal
+
+exec gosu postgres /opt/cpm/bin/start.sh "$@"
--- a/centos7/9.5/Dockerfile.postgres.centos7
+++ b/centos7/9.5/Dockerfile.postgres.centos7
@@ -36,6 +36,14 @@ RUN yum -y install postgresql95-server postgresql95-contrib postgresql95 \
 	pgbackrest \
  && yum -y clean all
 
+ RUN gpg --keyserver pool.sks-keyservers.net --recv-keys B42F6819007F00F88E364FD4036A9C25BF357DD4 \
+    && curl -o /usr/local/bin/gosu -SL "https://github.com/tianon/gosu/releases/download/1.7/gosu-amd64" \
+    && curl -o /usr/local/bin/gosu.asc -SL "https://github.com/tianon/gosu/releases/download/1.7/gosu-amd64.asc" \
+    && gpg --verify /usr/local/bin/gosu.asc \
+    && rm /usr/local/bin/gosu.asc \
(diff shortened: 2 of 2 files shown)


jmccormick2001:
hmmn, so far specifying fsGroup or supplementalGroups in the security context have allowed me to map to various file systems with the file permissions working and running the container as UID 26, this works for openshift too...I'm curious if you have tried either fsGroup or supplementalGroup with your gluster setup?  I have customers that don't want to run as root or install gosu so this is why I'm running the container as USER 26.

FaKod:
I personally do not have any issues with running a startup script as root, as long the process runs as a different user. Just like the "stock" postgres image does. But even if using root there shouldnt be a problem. If one needs that extra level of security, he should use rkt and it's virtualization layer.

Anyway, with the HEKETI standard setup and stock Kubernetes, neither fsGroup nor supplementalGroups work

jmccormick2001:
thanks, I'd really like this to work with heketi, I'll look into this one and evaluate it.

suchitpuri:
@jmccormick2001 i am facing the same issue, running the cluster on Kubernetes, the permissions of the /pgdata is messed up. I am using IBM Bluemix for kubernetes cluster and their nfs which i dont have much control on as the nfs. 

Let me know if there is a solution to fix this issue ( i tried using "runAsUser": 26, but it does not seem to work )

jmccormick2001:
can you provide me the permissions of the NFS mount you are trying to use?

suchitpuri:
I dont think, so i am using dynamic allocation of PV's in stateful set's. I just now tried with GCE as well and its giving the same permission error as in GCE the /pgdata is mounted by root and postgres user cant chown the directory

jmccormick2001:
I've been able to get it to work on GCE if you specify a fsGroup attribute in the security context....https://github.com/CrunchyData/crunchy-containers/blob/master/examples/kube/statefulset/set.json

take a look at that example of setting the fsGroup...would be curious if this works for you on GCE,

suchitpuri:
That really helped, i was looking at https://github.com/CrunchyData/crunchy-containers/blob/master/examples/kube/statefulset/set.json.dynamic 

At least i am past the stuck point.

jmccormick2001:
I spent some time this weekend working with Redhat's Container Native Gluster storage, I was able to get this to work using the container by a) specifying a GID when I create the Gluster volume (e.g. 590 as the GID), and b) by specifying a supplementalGroup value of 590 when running the container for example.

jmccormick2001:
retested this just now against heketi 4.0.0 and gluster 3.10, seems to work on kube 1.7.3, closing for now.

PR #174, opened 2017-06-26
Misc

xenophenes:
Most of the document referred to deprecated examples/documents, updated and filled in additional information so the document is more complete & accurate.

### DIFF
--- a/docs/examples.adoc
+++ b/docs/examples.adoc
@@ -1723,7 +1723,7 @@ kubectl scale pgset --replica=3
 ==== OpenShift
 
 This example shows how to use a StatefulSet (available
-in OpenShift Origin 1.5) to create a PostgreSQL cluster.
+in OpenShift Origin 3.5) to create a PostgreSQL cluster.
 
 Build the example by:
 ....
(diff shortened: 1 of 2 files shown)


jmccormick2001:
I went ahead and merged but did spot only a small thing to note....the Openshift versioning is almost always confusing to folks....Openshift Origin, the open source version, is at version 1.5.X, this is in line with the Kube version it is based upon (e.g. Kube 1.5.X).  However, the enterprise non-free version of Openshift now called Openshift Container Platform has a different version number, it is now at version 3.5 which implies it is based on Kube 1.5.

The reason for this version difference is historical, there was indeed an Openshift version 2.0 that was not based on Docker at all, so they kept that versioning scheme and when the Docker version of Openshift was announced they started it at version 3.0.  

So, in the documentation when we refer to Openshift we need to typically spell out both Openshift Origin 1.5 and Openshift Container Platform 3.5 as being sort of the same but different.  One is free and the other is not essentially but both are based on Kube 1.5.

xenophenes:
Interesting to note!

PR #180, opened 2017-07-06
yum improvements

tchughesiv:
its likely not perfect because i couldn't test the full build w/o your repo... but this should give you an idea of a preferred way to approach yum installs in rhel.  this will ensure only the proper host entitlements are being used during the install.

also, using latest rhel7 image in these changes and added openshift repo for nss_wrapper requirement.


jmccormick2001:
thank you we will review!

xenophenes:
Apologies for the length of time it's taken to review this. Would you mind fixing the conflict that's come up? I can take a look at this after that and test it out.

jmccormick2001:
thank you @tchughesiv, we will incorporate some of these changes into future releases.

Issue #178, opened 2017-06-30
upgrade rhel containers to latest RHEL version


jmccormick2001:
currently the rhel containers use rhel7.2, I would like to update those Dockeffiles to use rhel7.3.

jmccormick2001:
merged in changes from Sarah for this.

PR #186, opened 2017-07-14
For Issue #178

xenophenes:
Experiencing problems with laptop hard drive, so uploading my progress for now & charting where I'm at.

Support case has been submitted to Red Hat a few days ago to determine if there is a character limit for summary & description errors to avoid issues after implementing the detailed labels. I have typed out summaries & descriptions for each container in Google Docs, so once that answer is determined I'll be able to simply copy & paste and add that change.

Tested all containers with Docker, in progress with testing on OpenShift & Kubernetes. Specifically am testing RHEL7 images built with the latest RHEL version (7.3), with the code changes committed below.

Whilst running the master-replica-rc example for Kubernetes, I come across the error that the file replica-service.json doesn't exist. It looks like the run script attempts to create both the master-service.json and the replica-service.json files; master-service exists but replica-service does not. Any ideas on what happened there?


xenophenes:
Received confirmation from Red Hat in the support case I opened that there are no character limits for labels or annotations. Updated summary/description labels with descriptive text for RHEL and Centos Dockerfiles.

xenophenes:
When you can, please let me know thoughts on this:

> Whilst running the master-replica-rc example for Kubernetes, I come across the error that the file replica-service.json doesn't exist. It looks like the run script attempts to create both the master-service.json and the replica-service.json files; master-service exists but replica-service does not. Any ideas on what happened there?

I checked the history of that folder and it seems that entire file (replica-service.json) was deleted in https://github.com/CrunchyData/crunchy-containers/commit/8cbdedc08fa2d97a5d383cba7041bd1db65af52b. Was that an accident or intentional with the reference to it needing to be removed from the run script?

jmccormick2001:
I think it was removed by mistake....if you could re-add that service and
test
the example that would probably be the way to go for now.

thanks

PR #222, opened 2017-08-28
Make sure docker tag of new component matches others, for uniformity

crunchyjohn:
The new sim component doesn't match the others tag-wise, so this change will allow it to be uniform.


crunchyjohn:
Had some discussions-- I think this is good to go.

Issue #233, opened 2017-09-17
Conversion of terminology


xenophenes:
Issue: Usage of **master/slave** and **master/replica** in code and documentation.
Resolve: Switch to **primary/replica** across all code and documentation.

Purpose of raising this issue: To track the conversion of these terms in batches from start to finish. Will start with the conversion of all mentions of **slave**. This issue can be assigned to me.

Will also use this as an opportunity to revisit old documentation in addition to removing unnecessary files. General clean-up.

Will also need to verify correct terminology in Red Hat Connect product descriptions.

Evaluation:

- Term **slave** used 294 times across 95 files.
- Term **SLAVE** used 169 times across 53 files.
- Term **master** used 2250 times across 387 files.
- Term **MASTER** used 913 times across 157 files.

xenophenes:
**Question**:

These following directories seem to be underdeveloped or abandoned. Are they necessary to keep and continue developing, or can they be dropped?
- kubehelm/crunchy-postgresql-cluster/
- examples/portworx/
- docs/notes/

jmccormick2001:
done, all changes are merged into master.

PR #248, opened 2017-10-02
update pgpool relating to issue #233

xenophenes:


jmccormick2001 (changes requested):
the pgpool.conf file shows changes made to various pgpool configuration setting names...this is not possible as we don't own the pgpool codebase that will look for those setting names.

PR #261, opened 2017-10-06
Pg10 support

crunchyjohn:
Support for RHEL and CentOS on PG10.  I also made the Makefile a wee bit smarter.

Caveats:
1) The original branch had corresponding documentation changes to switch to PG10 as the default for the examples.  While that made sense in the branch itself for pg10, I did not think we were ready to update all of the docs to delegate PG10 as the default container option just yet.  If that assumption is incorrect, please let me know and I can put the changes back into the PR.

2) While the Red Hat "make all" target is error-free at the moment, the CentOS build can compile all pieces EXCEPT postgres-gis due to a missing dependency in PGDG, which is SFCGAL.  I have notified the yum.postgresql maintainer.  One that dependency is made available, the "make all" target should work for all CentOS / PGDG containers.


Makefile
```diff
@@ -2,12 +2,24 @@ ifndef CCPROOT
 	export CCPROOT=$(GOPATH)/src/github.com/crunchydata/crunchy-containers
 endif
 
+.PHONY:	default versiontest
+
+default:	all
```
abrightwell: Why not just move the 'all' target up to the top? `make` assumes the 'first' target is the 'default'.

crunchyjohn: @abrightwell Good suggestion.  I have made and pushed that change.

abrightwell (commented):
Generally speaking, would it make more sense to call it `10.0` instead of `10`? My thoughts are 1) convention and 2) organizing future versions, `10.1` for instance.

crunchyjohn:
@abrightwell Good question.  I thought about that, but the new PostgreSQL numbering convention for 10 drops a digit, and I think it makes sense for us to follow suit.  Any release in the 10 series (10.0 / 10.1, etc) would utilize the 10 Dockerfiles, similar to how any release in the 9.6 series (9.6.0 / 9.6.1, etc) utilizes the 9.6 Dockerfiles.  If we made a 10.0 folder, that's equivalent to having a 9.6.0 folder, which isn't really what is wanted here.  Generally, the containers project follows major releases only, so using "10" would follow that convention.

PR #264, opened 2017-10-09
Fixing run.sh so that it fails fasts, only rely on python and no othe…

jayunit100:
Some minor tweaks to make the run.sh smoother the first time around.... @jmccormick2001  @jconway ... 

This is part of my initial experimentation with this recipe, #265 .

### DIFF
--- a/examples/kube/master-deployment/cleanup.sh
+++ b/examples/kube/master-deployment/cleanup.sh
@@ -12,8 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-kubectl delete service master-dc
-kubectl delete service replica-dc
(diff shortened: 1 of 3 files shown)


jayunit100:
@jmccormick2001  does this look reasonable to you ?  Would be nice for our GKE deployments to not rely on non-standard linux binaries, this one just uses python and nothing else...

jmccormick2001:
maybe, I'll give it some thought, the change would be across many different example sets not just this example so the number of changes would be potentially large...also, we are going to add Helm charts for each kube example in the near term which changes things up as well, so this may be a non-issue if we go that route.  but I'll definitely keep it as a live issue to consider.

jmccormick2001:
fixed this another way using a different golang binary as a replacement for envsubst.

Issue #296, opened 2017-10-24
Liveness and Readiness checks should use PG_DATABASE


xqianwang:
In liveness.sh and readiness.sh, pg_isready function is using postgres database for checking liveness and readiness which is not accessible by PG_USER (application user).

PR #297, opened 2017-10-24
#296: Liveness and Readiness checks should use PG_DATABASE

xqianwang:


jmccormick2001:
thanks, but having the value default to 'postgres' make sense for almost all situations that I can think of, please elaborate on how this is useful and I'll reconsider..regards.

PR #334, opened 2017-11-21
fix link, add PG_PRIMARY_PORT

xenophenes:


xenophenes:
Testing something else, will reopen this PR with an additional commit added later.

PR #336, opened 2017-11-24
Add missing PG_PRIMARY_PORT, update all-in-one docs & script

xenophenes:


xenophenes:
Added a very minor installation instructions update. Everything in this is tested and ready to go, and fixes three examples. 

The only major change was "update all-in-one example" and replacing setup-env.sh in favor of all-in-one.

PR #340, opened 2017-12-08
Add Docker, OpenShift, and K8 examples for postgres-gis. Add entries to examples.adoc.

steve-hetzel:


jmccormick2001:
you might mention in the examples.adoc that 'psql \dx' output will show the postgis extensions have been loaded, this might not be clear to some users and also is a simple test for our purposes to document.  or perhaps some other postgis specific query as a simple example to show how this container is different than the normal postgres container.

steve-hetzel:
Added some additional commits with additional examples and typo/bug fixes.

jmccormick2001:
remove the Disney reference in the examples.adoc, in general don't refer to other companies in our documentation unless its a link to documentation...other than that change the PR looks good.

steve-hetzel:
Complete.

PR #352, opened 2017-12-20
Add oc process to postgis example for openshift

jasonodonnell:
Removing `expenv` from the openshift example and using the built in `oc process` to render the template.


steve-hetzel (commented):
comment removed

Issue #361, opened 2018-01-03
Master/Slave labels still exist


jasonodonnell:
In some examples we are still using `master/slave` notation:

[ms-primary](https://github.com/CrunchyData/crunchy-containers/search?utf8=%E2%9C%93&q=ms-primary&type=)
[ms-replica](https://github.com/CrunchyData/crunchy-containers/search?utf8=%E2%9C%93&q=ms-replica&type=)
[ms-replica-2](https://github.com/CrunchyData/crunchy-containers/search?utf8=%E2%9C%93&q=ms-replica-2&type=)
[m-s-rc-pvc-primary](https://github.com/CrunchyData/crunchy-containers/search?utf8=%E2%9C%93&q=m-s-rc-pvc-primary&type=)
[m-s-rc-pvc-replica](https://github.com/CrunchyData/crunchy-containers/search?utf8=%E2%9C%93&q=m-s-rc-pvc-replica&type=)
[m-s-rc-pvc](https://github.com/CrunchyData/crunchy-containers/search?utf8=%E2%9C%93&q=m-s-rc-pvc&type=)

jmccormick2001:
merged

Issue #360, opened 2018-01-02
docs update - XLOGDIR specified twice


jmccormick2001:
the XLOGDIR is mentioned twice in containers.adoc in the Postgres container section.

jmccormick2001:
merged

PR #364, opened 2018-01-03
Issue #360 + #361 in addition to more minor docs changes

xenophenes:


jmccormick2001:
can you find a Helm link instead of a bitnami link for this install help?  I'd prefer the  base source if possible instead of a separate vendor link.

xenophenes:
Done!

Issue #342, opened 2017-12-09
call pgbackrest via sshd container


jmccormick2001:
perform a backrest backup command using the sshd container sidecar

PR #395, opened 2018-01-22
Add SSHD to Postgres Containers

jasonodonnell:
This PR contains the feature requested in #342 to enable SSHD in the PostgreSQL container for DBA tasks (running pgBackrest) on the PostgreSQL and PostgreSQL-GIS images (RHEL/Cent).

I've added a new document sshd.adoc that contains details on configuring the containers to run SSHD.

A new example called postgres-sshd can be found in the docker, kube, and openshift example directories. The new volume made available in the PostgreSQL images was also added to the OpenShift pro example.

The Makefile targets weren't alphabetized so I ordered it.


thesteve0:
Adding sshd to a docker image is bad practice and generally frowned upon

Coming at you from my phone, please excuse typos, poor grammar, curtness,
and dashing good looks

jasonodonnell:
Agreed about the bad practice of adding SSHD to images, however, we're a bit tied here.  `pgBackrest` uses ssh as its communication protocol and is the only way to create dedicated backup hosts (without having to mount a unix socket to a shared volume).  This lays the foundation we need to build this architecture.

Issue #441, opened 2018-02-13
1.8.0 PITR Docker pg_xlog_replay_resume not found


altondinsmore:
xlog has changed in PG 10. It is now wal.

select pg_xlog_replay_resume();
select pg_wal_replay_resume();

https://www.dbrnd.com/2017/12/postgresql-10-important-name-change-from-xlog-to-wal-and-location-to-lsn/

jmccormick2001:
merged

Issue #440, opened 2018-02-12
1.8.0 PITR Docker Doc correction


altondinsmore:
Doc has...
'These segments are stored in the /tmp/primary-data/master-wal directory.'

Directory does not exist....

jmccormick2001:
merged into master

Issue #439, opened 2018-02-12
1.8.0 PITR Docker function pg_xlog_replay_resume() does not exist


altondinsmore:
Missing function....
psql -h 127.0.0.1 -p 12001 -U postgres postgres -c 'select pg_xlog_replay_resume()'
Password for user postgres: 
ERROR:  function pg_xlog_replay_resume() does not exist
LINE 1: select pg_xlog_replay_resume()
               ^
HINT:  No function matches the given name and argument types. You might need to add explicit type casts.

jmccormick2001:
merged into master

Issue #420, opened 2018-02-06
All examples should use CCP_IMAGE_PREFIX


xenophenes:
There are some examples still using crunchydata/ as hard code rather than using the CCP_IMAGE_PREFIX variable. This should also be replaced in the /bin/pull-from-dockerhub.sh and /bin/push-to-dockerhub.sh scripts.

You can assign this to me if you want!

jmccormick2001:
merged into master

PR #452, opened 2018-02-16
Issues #420, #440, #441, #439

xenophenes:
- replaced hard coded crunchydata/ in all run scripts (pull-from-dockerhub.sh, push-from-dockerhub.sh, etc)
- replaced hard coded crunchydata/ in all JSON files
- added arguments to processing commands (such as oc process -f) to process that variable
- tested the following examples in the following environments after changes:

https://docs.google.com/spreadsheets/d/1iG8nHZUDbO91K-sF6HUuM0CDLxbf62rva2c61i-2-JY/edit?usp=sharing

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -39,17 +39,17 @@ docbuild:
 
 backrestrestore: versiontest
 	docker build -t crunchy-backrest-restore -f $(CCP_BASEOS)/$(CCP_PGVERSION)/Dockerfile.backrest-restore.$(CCP_BASEOS) .
-	docker tag crunchy-backrest-restore crunchydata/crunchy-backrest-restore:$(CCP_BASEOS)-$(CCP_PG_FULLVERSION)-$(CCP_VERSION)
+	docker tag crunchy-backrest-restore $(CCP_IMAGE_PREFIX)/crunchy-backrest-restore:$(CCP_BASEOS)-$(CCP_PG_FULLVERSION)-$(CCP_VERSION)
 
 backup:	versiontest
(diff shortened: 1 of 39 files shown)


xenophenes:
Issues #439 #440 #441 all taken care of with three additional commits to this branch. 

- Locations updated from /pg_xlog to /pg_wal
- Functions changed from pg_xlog_replay_resume(); -> pg_wal_replay_resume();
- run-restore-pitr.sh was referencing incorrect wal directory, changed reference
- Docs were referencing wrong directory for checking wal, changed reference
- Added example output where it was indicated but not present to the docs

Tested on Docker with rebuilt images on a clean machine. Double checked the process and all logs.

abrightwell (commented):
Seems that we'd probably want to check the version first and use `wal` vs `xlog` based on that. Otherwise, pre-pg10 versions will no longer work.

I suppose this could be accomplished a couple of different ways:

1) If you have access to the data directory (which looks like most of these cases do), then checking `PG_VERSION`.

2) `psql -c "SELECT version();"`

xenophenes:
OK; I'll take a look at updating these commits to take that into account.

xenophenes:
Taken into account. Thanks @davecramer for the helpful input!

jmccormick2001:
please pull out the simple stuff from the PR like CCP_IMAGE_PREFIX changes from the xlog/wal changes into separate PRs so I can better review/test them, they seem to be 2 different Issues here.  thanks.

PR #457, opened 2018-03-01
Updating some Copyright dates

crunchyheath:


davecramer:
I don't think it is necessary to update copyright dates. My understanding of how this works is that anything that existed in 2015 can retain that date. Any new files would have new dates on them.

PR #462, opened 2018-03-02
Added PVC to the primary-pvc example template for OpenShift.

jsm84:


jmccormick2001:
thanks Josh, I'm heavily working on the storage configuration, this helps.  The new storage configuration will touch all examples (kube and Openshift) and each example will have its own PVCs like this one...users will have the up front option of specifying either hostpath, nfs, or a dynamic storage class as their storage config, then each example will take that into account.  This will be added over probably a few different releases starting with 1.8.2.

PR #478, opened 2018-03-14
issue #420 replace hardcoded crunchydata with CCP_IMAGE_PREFIX and other fixes

xenophenes:
- updated /mnt/nfsfileshare in pgpool run script to reflect $PV_PATH instead
- updated primary-restore and docker restore examples to use primary-pvc-backups instead of basic-backups
- replaced all instances of hardcoded crunchydata/ with $CCP_IMAGE_PREFIX
- tested affected files


examples/openshift/workshop/watch.json
```diff
@@ -14,7 +14,14 @@
         "name": "CCP_IMAGE_TAG",
         "description": "the image tag to use"
     }, {
+<<<<<<< HEAD
```
davecramer: @xenophenes there is a conflict here which needs to be resolved first

xenophenes: Thanks! Resolved.

PR #632, opened 2018-05-24
Fix bad test about non existence of directory

JrCs:
The test about the non existence of a directory is invert.
Command like:
```bash
pgo upgrade mycluster --upgrade-type=major --ccp-image-tag=centos7-10.4-1.8.3
```
can't work without this fix.

### DIFF
--- a/bin/common/common_lib.sh
+++ b/bin/common/common_lib.sh
@@ -60,7 +60,7 @@ function env_check_info() {
 }
 
 function dir_check_err() {
-    if [[ -d ${!1} ]]
+    if [[ ! -d ${!1} ]]
     then
         echo_err "The $1 directory does not exist and is required."
         exit 1


bin/common/common_lib.sh
```diff
@@ -60,7 +60,7 @@ function env_check_info() {
 }
 
 function dir_check_err() {
-    if [[ -d ${!1} ]]
+    if [[ ! -d ${!1} ]]
```
xenophenes, line 63: Thanks very much for catching this! It is indeed incorrectly coded. However, the correction you made is still incorrect - after testing to be certain, it should be:

`if [[ ! -d ${1} ]]`

xenophenes (changes requested):
Once the comment I made is addressed and the commit re-submitted, this will be good to approve & merge.

bin/common/common_lib.sh
JrCs: No because ${!1} is use to dereference the variable name pass in the argument. So my fix is correct.

xenophenes: My apologies, I was wrong. After revisiting the test I had originally used I found I didn't use variable names to test with but rather hardcoded directory paths. Re-tested using variable names and it is indeed correct. @jmccormick2001 this is good to go

Issue #582, opened 2018-04-19
Integrate pgmonitor into metrics stack


jasonodonnell:
**I'm submitting a ...**

 
 - [X] feature request

**What is the motivation or use case for changing the behavior?**
pgMonitor offers a lot more metrics and better dashboards.

pmereddy1:
pg_stat_activity_count metrics exported by collect container are not accurate.
You will see only one active session regardless of the activity on the postgres cluster.
Upon investigation, I learned that collect container is using primaryuser to gather information on session activity and this db user is not authorized to check session activity. Thanks Jason.

jmccormick2001:
please see Keith Fiske for their 2.0 tag soon to land, if possible, integrate that new version.

jasonodonnell:
@jmccormick2001 Spoke with Keith, 2.0 uses the newest version of Grafana (5).  You cool with me updating to that version?

jmccormick2001:
yep, lets upgrade.

PR #713, opened 2018-07-30
Add pgMonitor integration

jasonodonnell:
This pull request satisfies the requirements in #582 by adding pgMonitor into the metrics stack.

Here's what can be found in this PR:

* pgMonitor integration - during build we checkout pgMonitor repo and add the files we need for pgMonitor.  The `install_deps.sh` checks out the commit with the changes required for pgMonitor to work with containers.

* New versions of software:
  * Grafana 5.2.1
  * Prometheus 2.3.1
  * PostgreSQL Exporter 0.4.6
  * Node Exporter 0.16.0

* Prometheus and Collect now work with replicas (previously it didn't).

* Fixed a bug where Prometheus would contain old deployments.  This is done by changing the labels in prometheus to make them not so unique.

* Applies specific version functions by detecting the PostgreSQL version.

* PostgreSQL container has a new environment variable `PGMONITOR_PASSWORD` - this creates a new role `ccp_monitoring` to allow the collect container to scrape metrics using SECURITY DEFINER functions.  This is required to allow collect to get the info it needs without requiring a superuser role.

* Removed API calls for registering datasources and dashboards now that Grafana 5 supports more automated deployments using configuration files.


jasonodonnell:
Develop was missing a commit that's currently in master (https://github.com/CrunchyData/crunchy-containers/commit/98d018491b2decb2fc8b92dd61150a1f9976a375#diff-32aad64d375bdd71441f66e0e7840403), this PR picked up that missed commit.

PR #729, opened 2018-08-08
Fix http redirect

scrummyin:
Pdbadger was sending a 301 so that it would not regenerate the
documentation when hit again in a browser. Instead, the browser would use
their cache and just bring users to static. Switching from 301 to 302
fixes this issue.

**Checklist:**

 
**Type of Changes:**
 
Currently, in a given browser, you will only be able to generate the pgbadger report once, as a local cache will cause it to skip that step.

You can access /api/badgergenerate again and agian to regenerate the report and it will not be cached.

**Other information**:
I have a screen capture showing this behavior if you would like, but it's an image and I was not going to upload unless requested.

### DIFF
--- a/badger/badgerserver.go
+++ b/badger/badgerserver.go
@@ -53,5 +53,5 @@ func BadgerGenerate(w http.ResponseWriter, r *http.Request) {
 	}
 
 	log.Println("Report generated.  Redirecting..")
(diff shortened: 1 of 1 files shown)


jasonodonnell (approved):
Works as advertised.  Thanks for this patch @scrummyin !

PR #732, opened 2018-08-08
Add handler for / in pgbadger http server

scrummyin:
Currently you have to remember /api/badgergenerate which is a longer url
than I would like to keep in my head. I added a root handler with two
code paths. If /report/index.html does not exist you get a redirect to
/api/badgergenerate. If it does exist you get a redirect to /static. I
don't believe this change will break anything because there is nothing
listening there currently on this http server.

**Checklist:**

 
**Type of Changes:**

 
The pgbadger http server does not have a handler on /

There is now an intelligent handler on /

**Other information**:
This is a feature add, please let me know if you would prefer this work a different way.


jasonodonnell (approved):
Tested and worked as expected.  This is a great change @scrummyin, thanks for the patch.

PR #767, opened 2018-09-27
Correcting Docker docker-vg Instructions

dtseiler:
We shouldn't be using fdisk or vgcreate manually prior to running docker-storage-setup. Doing so causes docker-storage-setup to complain and exit. We also need to make sure they remove existing STORAGE_DRIVER settings in the setup file. Old instructions said to "add" the lines, meaning if the user left STORAGE_DRIVER="overlay2" in place (it's there on CentOS7) then their storage will be screwed up.

**Checklist:**

 
**Type of Changes:**

This is a documentation correction.


jmccormick2001:
I've updated that docs section due to other issues I saw with that section, thanks for reporting this.

PR #773, opened 2018-10-02
Remove godep, add dep and vendor code

jasonodonnell:
This pull request removes `godep` and replaces it with `dep` (#757).  All the vendor code is being checked in with this update.

Additionally I fixed:

* Updated go build dockerfiles to not use godep (no longer required since vendor code is checked in)
* Job method changed in kubeapi for the `scheduler`, so updated test-harness to work with new code


jmccormick2001:
need dep installed in the install-dep.sh script

PR #814, opened 2018-10-17
Fix run script in OCP templates

jasonodonnell:
Old `create-all.sh` script didn't work with templates.


jasonodonnell:
Closes crunchydata/crunchy-containers-test#115

PR #906, opened 2018-11-14
Update statefulset SA rbac

jasonodonnell:
The Statefulset RBAC was a `cluster-admin` account, when all it really needs is the ability to update pod labels.  I've limited the account to just what it needed and additionally split out the RBAC into a separate script so you can run it as cluster administrator.

Closes CrunchyData/crunchy-containers-test#165.
Closes CrunchyData/crunchy-containers-test#167.

### DIFF
--- a/examples/kube/statefulset/cleanup.sh
+++ b/examples/kube/statefulset/cleanup.sh
@@ -16,8 +16,7 @@ source ${CCPROOT}/examples/common.sh
 echo_info "Cleaning up.."
 
 ${CCP_CLI?} delete --namespace=${CCP_NAMESPACE?} statefulset statefulset
-${CCP_CLI?} delete --namespace=${CCP_NAMESPACE?} sa statefulset-sa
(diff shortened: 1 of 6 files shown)


andrewlecuyer (commented):
@jasonodonnell, what if we change the **clusterrole** to a **role**, and the **clusterrolebinding** to a **rolebining**?  This would allow an OCP user to run the example themselves (i.e. without an admin), since they will have been automatically assigned to the **admin** role for the project  they created to run the examples (e.g. the **demo** project), which will allow them to manage **roles** and **rolebindings** within that **project/namespace**.  I test this locally, and was able to run both **run-rbac.sh** and **run.sh** successfully, and then succesdfuly test the **statefulset** example without required any further  administrative access.

Also, just a quick note that the following command is in both **run-rbac.sh** and **run.sh**:
```
expenv -f $DIR/statefulset-sa.json | ${CCP_CLI?} create --namespace=${CCP_NAMESPACE?} -f -
```

jasonodonnell:
Yeah agreed.  Making the change.  Good catch on the duplicate command.

PR #917, opened 2018-11-21
Fix kubeapi bug with deployment labels

jasonodonnell:
The kubeapi package was iterating over each label in a deployment individually which caused issues when generating a list of pods with any of those labels.

### DIFF
--- a/tools/kubeapi/kubeapi.go
+++ b/tools/kubeapi/kubeapi.go
@@ -1,14 +1,16 @@
 package kubeapi
 
 import (
+	"bytes"
+	"fmt"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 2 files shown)


tools/kubeapi/kubeapi.go
```diff
@@ -1,14 +1,16 @@
...
 
 import (
+	"bytes"
+	"fmt"
 	"os"
 	"path/filepath"
 	"time"
 
 	"k8s.io/client-go/kubernetes"
+	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
```
scalaenthusiast: Why was this line added?

Issue #653, opened 2018-06-29
PostgreSQL readiness probe race condition


jasonodonnell:
**I'm submitting a ...**

 
 - [X] bug report
 - [ ] feature request
 - [ ] support request

The [PostgreSQL readiness probe](https://github.com/CrunchyData/crunchy-containers/blob/master/bin/postgres/readiness.sh) can prematurely tell Kubernetes that the container is ready to use.  The database is temporarily shut down after initialization which can lead to a race condition where the probe was successful but the database is off or starting up again.

We should refactor this probe to look for better signs that the startup was successful such as: 
* presence of custom databases/roles
* installed extensions
* timestamp inserted into a metadata table after the db has started

PR #925, opened 2018-12-05
[crunchy-postgres] Limit listener of temporary pg to unix domain socket

gkranis:
**Checklist:**

 
**Type of Changes:**

 
- The replica node uses `waitforpg` to detect when the master is ready for connections.
- Before the real master node comes up, a temporary master is used  in `initialize_primary`.
- There's a chance that `waitforpg`, if fast enough, will connect to the temporary master node instead of the real one. 

Same issue as described here https://github.com/CrunchyData/crunchy-containers/issues/653

- Temporary primary listens only on unix domain socket (same approach that the official pg docker images [use](https://github.com/docker-library/postgres/blob/040949af1595f49f2242f6d1f9c42fb042b3eaed/10/docker-entrypoint.sh#L114-L116)).
- `pg_isready` in `initialize_primary` probes the unix domain socket instead of the ip
- The rest of `psql` commands in `initialize_primary` already used the unix domain socket so no change was needed
- `waitforpg` can't connect any more to the temporary primary

**Other information**:

Tested by patching those changes into `crunchydata/crunchy-postgres:centos7-10.5-2.1.0`. We just use a cluster of 2 initially empty crunchy-postgres in our CI run. Works for us :)

### DIFF
--- a/bin/postgres/start.sh
+++ b/bin/postgres/start.sh
@@ -304,13 +304,12 @@ function initialize_primary() {
         echo "Starting database.." >> /tmp/start-db.log
 
         echo_info "Temporarily starting database to run setup.sql.."
-        pg_ctl -D $PGDATA start
+        pg_ctl -D $PGDATA -o "-c listen_addresses=''" start
 
         echo_info "Waiting for PostgreSQL to start.."
         while true; do
             pg_isready \
-            --port=$PG_PRIMARY_PORT \
-            --host=$HOSTNAME \
+            --host=/tmp \
             --username=$PG_PRIMARY_USER \
             --timeout=2
             if [ $? -eq 0 ]; then


jasonodonnell (changes requested):
Hi @gkranis,

Thanks for the patch, I have tested this and it works as expected.  I think it would make a good addition.

You will need to resubmit this pull request to the `develop` branch before we can approve this.

gkranis:
@jasonodonnell done

longtimeago:
@jasonodonnell would love to see the release with this fix. Thanks! :)

jasonodonnell:
@longtimeago This will be included in our 2.3.0 release which is scheduled for mid January.  Thanks!

PR #1015, opened 2019-02-28
Add local login for postgres user to pgpool

flamingdumpster:
**Checklist:**

**Type of Changes:**

 
 - [ x  ] New feature (non-breaking change which adds functionality)

Default configuration of pgpool container does not allow users to login from localhost (from within the pgpool container)

inside container, pgpool can be logged into with postgres and a default password. Uses same pasword as pgpool example

**Other information**:


jasonodonnell (changes requested):
I don't think we should be setting a default `pool_passwd` file as this will always depend on your deployment (currently this assumes your password is `password`).

This file should be mounted in the examples.  

Thoughts @flamingdumpster @jmccormick2001 ?

jmccormick2001:
${CCP_CLI?} create --namespace=${CCP_NAMESPACE?} configmap pgpool-pgconf \
        --from-file=$DIR/configs/pgpool.conf \
        --from-file=hba=$DIR/configs/pool_hba.conf \
        --from-file=psw=$DIR/configs/pool_passwd

it might just be me but I'm not seeing where that configmap is used...maybe cruft leftover from way back?  The Secret is used to store the pgpool config files instead it appears.  Please verify that the configmap can be removed from the example.


I'm ok with the pgpool password being 'password' for this example, I think its always been that?  The example has always worked against the primary-replica and has assumed they have passwords of 'password' too if I recall, those being hardcoded in the pgpool.conf file.  For a real deployment these passwords would indeed need to be configurable.  So, I'm probably missing the point you are making Jason.  Lets chat tomorrow if I'm totally missing the boat on this.

jasonodonnell:
Agreed with the configmap being unused - the secret is the important part in that example and so the configmap could be removed.

What I'm pointing out here is the passwd file included in this PR is part of the image when this file should only be injected during runtime (like our example does).

jmccormick2001:
ah, ok, yes, I'm looking at the example and not the image!  ok, yes, the pool_passwd file should not be in the image itself since its mounted by the examples in the Secret...

flamingdumpster:
Ok, I'll tweak this in the AM

jasonodonnell (approved):
Changes look good 👍

jmccormick2001:
lets see if the blank pool_passwd file in /conf/pgpool can be removed.

PR #1022, opened 2019-03-04
Doc update for install and build

thesteve0:
ok doc finally in without the fedora pieces. @cahoonpwork


thesteve0:
Forgot to branch from the developer branch. One of these days I will get it. New PR incoming

PR #1025, opened 2019-03-05
Add pgBench image

jasonodonnell:
This adds a new image `crunchy-pgbench`, which can be used to run benchmarks against PostgreSQL clusters.

Features include:
* Environment variables to tune `pgbench` such as scale, number of clients, number of transactions per client.
* Custom flags for both pgbench initialization and the benchmark
* Mount a custom `transaction.sql` file to `/pgconf` for advanced benchmarking use cases

Removed the remaining parts of `crunchy-sim` as this is no longer needed with this new image.

[CH717]


tjmoore4 (commented):
Successfully reviewed.

PR #1031, opened 2019-03-28
Remove 'yum update' Command from crunchy-postgres-gis Dockerfiles

andrewlecuyer:
Removed the `yum update` command from all **crunchy-postgres-gis** Dockerfiles.  This will ensure the version of pgBackRest installed on the **crunchy-postgres** base image is maintained in the **crunchy-postgres-gis** container, preventing the potential for version mismatch errors when used with PGO.

[ch3109]

**Checklist:**

 
**Type of Changes:**

 
The pgBackRest version could be updated while building a **crunchy-postgres-gis** container, which can subsequently result in pgBackRest version mismatch errors with PGO.

A `yum update` command is no longer run when building a **crunchy-postgres-gis** container, preventing pgBackRest from upgrading to a version that is incompatible with PGO.

**Other information**:

### DIFF
--- a/centos7/10/Dockerfile.postgres-gis.centos7
+++ b/centos7/10/Dockerfile.postgres-gis.centos7
@@ -16,8 +16,7 @@ LABEL name="crunchydata/postgres-gis" \
 
 USER 0
 
-RUN yum -y update \
- && yum -y install \
+RUN yum -y install \
     R-core libRmath plr10 \
     postgis24_10 postgis24_10-client \
  && yum -y clean all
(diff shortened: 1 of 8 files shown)


thesteve0:
The problem is that if we do this then we could miss critical security
updates. Can we not just 'yum install' after  doing the yum update?

andrewlecuyer:
@thesteve0, being that the crunchy-postgres-gis container is built of off the crunchy-postgres  container, which does still have the `yum update` command, any necessary package updates (e.g. security updates) would be inherited by the crunchy-postgres-gis container.  This means in order to do a full rebuild of a crunchy-postgres-gis container, the crunchy-postgres container should/would first be built.  This ensures the crunchy-postgres-gis container inherits any necessary package updates, while also keeping package version across the  crunchy-postgres-gis and crunchy-postgres containers in sync.

PR #1059, opened 2019-05-15
Remove pgbouncer module from postgres containers 

flamingdumpster:
**Type of Changes:**

 
 - [x ] Bug fix (non-breaking change which fixes an issue)

Functions as designed

No change in container behavior. Removing pgbouncer module (a short term fix for operator) which is no longer needed. Feature has been implemented in operator that provides this functionality. 

**Other information**:


jasonodonnell (changes requested):
These modules should not be removed from the Postgres container.  As I understand it this code is duplicated in Operator, however, not everyone uses operator and this is a very breaking change to people who are using these containers together.  This code is what sets up the auth_query user and is required if you want to use our pgbouncer container as is.

I don't see why this can't exist in both places - can you think of any reason why?

flamingdumpster:
Yeah, I missed some pieces on this. There are suppose to be scripts added to the example. Will update with those pieces.

jasonodonnell:
This looks better but there's a couple of things I'm concerned about with this:

1. This is to be run as a superuser, we should not be encouraging people to use superuser accounts remotely.  I think this could be solved by the second point.
2. This assumes its being run in a cluster (you're hitting the service IP, most users are not running scripts on say a kube master host).  This might be better as a mounted post-start script the postgres container.
3. pgBouncer test in the harness isn't updated to reflect this change.

flamingdumpster:
Will update as suggested in 2, which should take care of 1 and 3.

PR #1062, opened 2019-05-20
Add node-exporter image

jasonodonnell:
Removed node-exporter from the collect image so it can be run as a daemonset.

[CH701]
[CH3652]


flamingdumpster (commented):
No comments. Looks good from what I can tell.

PR #1112, opened 2019-09-05
update to the latest pgbackrest version

lkhomenk:
2.16 version has crucial fixes related to S3 backups
https://pgbackrest.org/release.html
* Retry S3 RequestTimeTooSkewed errors instead of immediately terminating

**Checklist:**

 
**Type of Changes:**

 
S3 backup fails to often due to S3 reply-response time being to big. which led to zombifying pgbackrest. which required backrest-repo pod restart

New exception can be handled right way with retry

**Other information**:
Contributing guide is hidden on access.crunchydata.com which needs registration, right?

### DIFF
--- a/ubi7/11/Dockerfile.postgres.ubi7
+++ b/ubi7/11/Dockerfile.postgres.ubi7
@@ -24,7 +24,7 @@ COPY conf/atomic/postgres/help.md /help.md
 COPY conf/licenses /licenses
 COPY licenses /licenses
 
-ENV PGVERSION="11" BACKREST_VERSION="2.15.1"
+ENV PGVERSION="11" BACKREST_VERSION="2.16"
 
 # if you ever need to install package docs inside the container, uncomment
 #RUN sed -i '/nodocs/d' /etc/yum.conf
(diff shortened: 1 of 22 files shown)


ubi7/11/Dockerfile.postgres.ubi7
```diff
@@ -34,7 +34,7 @@ ADD conf/RPM-GPG-KEY-crunchydata  /
 ADD conf/crunchypg11.repo /etc/yum.repos.d/
 RUN rpm --import RPM-GPG-KEY-crunchydata
 
-RUN rpm -Uvh https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm 
+RUN rpm -Uvh https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm
```
lkhomenk, line 37: I guess, this was "linter" fix

lkhomenk:
2.17 is just arrived, not sure about it's updates. but this fix from 2.16 is needed 1000%

cahoonpwork:
Thanks, let me look into 2.17 and see if we should go ahead and move to that.  If so I have a script I can run to update the version.  We have to update the Operator backrest version too to keep them in sync.

cahoonpwork:
Updating to 2.17, will have that merged in this morning.

PR #1120, opened 2019-09-20
Update to pgMonitor v3.2 & Remove node_exporter container [ch4549][ch5698]

jkatz:
**Checklist:**

 
**Type of Changes:**

 
 - [x] Breaking change (fix or feature that would cause existing functionality to change)

Currently pgMonitor ~2.0 is being deployed. The node_exporter sits as its own sidecar, but is still referenced from the crunchy-collect documentation.

## Updated pgMonitor dependencies and compatibility.

This brings the container suite inline with the requirements for
pgMonitor v3.2, with a few differences noted below.

The following versions are now utilized:

- pgMonitor 3.2
- Prometheus 2.9.2
- Grafana 6.3.4

The postgres_exporter is still pegged at 0.4.7.

This commit does not utilize the OS, etcd, or Prometheus dashboards
that pgMonitor makes available as we do not ship the node_exporter
(which would not provide accurate metrics as to how a PostgreSQL
container is affecting a system) or prefer to let the user decide
how they wish to monitor additional services.

## Remove node_exporter from the container suite.

node_exporter would return metrics for the entire host machine, instead of
the specific container that it was a sidecar for. This removes node_exporter
from the container suite completely.

**Other information**:

Removing `node_exporter` could be a breaking change in some environments, but it's presently not providing those environments much favors.

I broke the changes out as two separate commits so they can be applied independently.


jkatz:
@andrewlecuyer I've rebased my patch to remove the conflict.

The UBI7 `node_exporter` container had a modification in a different commit, whereas this patch removes `node_exporter` completely.

tjmoore4 (approved):
Built and tested images in both Centos and RHEL environment. Worked as expected and no build errors were encountered.

PR #1148, opened 2019-11-19
refactors base image from builds

wilybrace:
A bulk of the fresh build of containers involves performing a full
yum update across every container image which is an unnecessary
time expense for the same intended state. Further, running N
separate yum updates runs a potential risk of either network failures
or package updates making later image builds have a different OS
base from the others.

This change creates a crunchy-base and crunchy-pg-base image from
which all other images base themselves, ensuring consistency of
the yum update set and expediting the build.

Quality of life changes as part of this change:
  * Parameterized release version
  * Parameterized prefix and base version for non base images
  * Consolidated image building rules to reduce copypasta
  * Docker/Buildah switch based on variable
  * Reordering some steps for increased cache hits
  * Disable ignoring partial yum installation failures
  * Convert pgbadger and scheduler builds to multi-stage build

This commit also flattens the Dockerfile structure

After consolidating settings into base images and parameters, there
are few differences between the Dockerfiles for different PG versions
This commit moves all pg-based Dockerfiles into a single folder,
parameterizing any differences other than the upgrade container.

WIP commits were squashed for simplicity of resolving rebase

As of result of these changes, the build runs approximately 50-70% faster for a full build of all container images.

**Checklist:**

 
 - [X] Have you updated or added documentation for the change, as applicable?
 - [X] Have you tested your changes on all related environments with successful results, as applicable?

**Type of Changes:**

 

centos7/Dockerfile.base.centos7
```diff
@@ -0,0 +1,21 @@
+ARG RELVER
+FROM centos:7
```
abrightwell, line 2: Would it be better to use a more specific tag here?  For instance `centos7:centos7.7.1908`?

jkatz: cc @crunchyjohn @crunchyheath 

From my perspective, it's one more thing to maintain and if we are always getting the latest and greatest CentOS 7 I'd prefer that. However, if we had to peg it at a particular version due to, say avoid an upgrade for a breaking reason, I'd would not be opposed.

However, open to other ideas around this.

abrightwell: Yeah, my thoughts are around configuration management and reproducible builds. Luckily, with this refactor, it isolates management of it to a single file.

wilybrace: It is now only in one place (base) and the metadata tags the os version as 7.7, so I'd either drop the os metadata for maintenance or be fine with upkeeping both the metadata and specific tag

abrightwell: I have no strong opinions on the metadata other than making sure that `docker inspect` can reliably produce the specific version (and other image identification/verification) information.

jkatz (changes requested):

Makefile
```diff
@@ -3,218 +3,288 @@ ifndef CCPROOT
...
 # Default values if not already set
-CCP_PG_VERSION?=11
-CCP_PG_FULLVERSION?=11.6
-CCP_PATRONI_VERSION?=1.5.6
-CCP_BACKREST_VERSION?=2.18
-
+CCP_BASEOS ?= centos7
+CCP_PG_VERSION ?= 12
+CCP_PG_FULLVERSION ?= 12.0
+CCP_PATRONI_VERSION ?= 1.5.6
```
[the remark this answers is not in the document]
wilybrace: I'll update to the latest, I think I set these values around Thursday last week...

wilybrace:
The changes in this PR happen to allow some use of the `-j` flag in Make for concurrent building. My experiments suggest another 50% reduction in build time from doing so, but tracing problems becomes crazy difficult. Future Makefile changes will better support a blind usage of `-j` without losing troubleshooting. For now, YMMV:
```
make cc-pg-base-image
make -j4 pgimages extras >/dev/null
```
The above invocation ensures the base images are available for the concurrent images and piping stdout to /dev/null allows seeing stderr lines, should they occur.

wilybrace:
Pending:
- any further review comments / feedback
- successful test runs from the build team on the rhel7 / ubi7 dockerfile sets.

jkatz (approved):
Looked through the changes. Performed a full CentOS 7 build successfully. Looks good.

crunchyheath:
I agree all of the CentOS PG9.5, 9.6, 10, 11, and 12 all pass.  However only the PG10, 11, and 12 RHEL builds succeed.  Both the PG9.5 and 9.6 builds fail with this error:

No package postgis25_96 available.
21:42:14  Error: Not tolerating missing names on install, stopping.
21:42:14  error building at STEP "RUN yum -y install 		--enablerepo="epel,rhel-7-server-optional-rpms" 		--setopt=skip_missing_names_on_install=False 		libRmath 		pgrouting_${PG_MAJOR//.} 		plr${PG_MAJOR//.} 		postgis${POSTGIS_LBL}_${PG_MAJOR//.} 		postgis${POSTGIS_LBL}_${PG_MAJOR//.}-client 		R-core 		texinfo-tex 		texlive-epsf 	&& yum -y clean all --enablerepo="epel,rhel-7-server-optional-rpms"": error while running runtime: exit status 1
21:42:14  make: *** [postgres-gis-pgimg-build] Error 1 

and missing postgis25_95 for the PG95 build

crunchyheath:
Using the CCP_POSTGIS_VERSION build variable I am able to build all of the RHEL and CentOS versions of the container suite with no issues on this branch.  So I you have my +1 for this to go in.  As a data point these changes seem to cut the build times down by about 20 minutes for the RHEL builds and between 20-40 minutes for the CentOS builds.  This is with a single build data point, but they seem to be pretty consistent.

crunchyheath (approved):

PR #1150, opened 2019-11-25
Add Automated Tests for Postgres HA and Postgres GIS HA Examples

tjmoore4:
**Checklist:**

 
**Type of Changes:**

 
New containers, Postgres HA and Postgres GIS HA have been added to the 
Crunchy Containers suite and need automated tests as already exist
for the other containers.

In keeping with the existing integration test suite, new tests are
added for the Postgres HA and Postgres GIS HA container examples.
These tests run the previously developed examples that deploy two
pods with the Patroni HA solution configured. The tests will run as
follows:

-In a new namespace, check that the relevant deployments and pods are
  created and ready
-Test the database connectivity on the first pod
(-For the GIS example, verify that the postgis extension is installed)
-Return the Patroni status information and validate the pods show as
  running and there is exactly one leader assigned to the Patroni cluster
-Cleanup the created resources

This ensures that the basic operation of the relevant examples are
working as expected.

**Other information**:
Please note: These tests do not currently work correction in an Openshift 
environment due to OCP's method of dynamically assigning an allowed User ID 
range when a new OCP project (aka namespace) is created. Because these tests are
designed to run in their own namespace, a more robust solution will need to be 
implemented in the future.

Issue: [ch6189]

### DIFF
--- a/tools/test-harness/postgres_gis_ha_test.go
+++ b/tools/test-harness/postgres_gis_ha_test.go
@@ -1,0 +1,172 @@
+package tests
+
+import (
+	"testing"
+	"encoding/json"
+)
+
+// Test the Postgres HA container by running the relevant example and validated the
+// various resources are created and running with the correct statuses
+func TestPostgresGISHA(t *testing.T) {
+	t.Parallel()
+	t.Log("Testing the 'postgres-gis-ha' example...")
+	harness := setup(t, timeout, true)
+	var pods []string
+	env := []string{}
+
+	t.Run("Run the Postgres HA container example", func(t *testing.T) {
+		_, err := harness.runExample("examples/kube/postgres-gis-ha/run.sh", env, t)
+		if err != nil {
+			t.Fatalf("Could not run example: %s", err)
+		}
+	})
+
+	// Validate that deployments are created, that relevant pods are created for those deployments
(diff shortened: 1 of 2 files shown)


jkatz (approved):

tools/test-harness/postgres_gis_ha_test.go
```diff
@@ -0,0 +1,172 @@
...
+		if len(extensions) < 1 {
+			t.Fatalf("extensions less then 1, it shouldn't be: %d", len(extensions))
+		} else {
+			t.Logf("Number of extensions: %v",len(extensions))
+		}
+
+		// Search through the installed extensions. If postgis is not installed
+		// the test fails
+		foundGIS := false
+		for _, extension := range extensions {
```
thesteve0, line 97: This has bitten us before where this passes but we don't exercise the extension to see if it actually works. Should we be doing something a bit more "real" here

jkatz: @thesteve0 We could do a bit more. In one of the tests we have, we call `SELECT postgis_full_version();` and that could be applied here as well.

Please feel free to submit a patch :wink: 

(We should have one shortly, actually)

PR #1157, opened 2019-12-05
[ch6458] Ensure pgAdmin4 displays the actual version being used

jkatz:

**Type of Changes:**

 
Previously the entire config file for pgAdmin4 was being overwritten by
our own local settings.

We only need to overwrite a few settings for
this container to work, so the minimum overwritable settings are now
applied.

This also fixes an issue where the directory for logging may not be
present as well, which would lead to a failure at container startup.

**Other information**:

I was unable to test the RHEL / UBI builds. The actual run of the CentOS 7 builds had some issues due to current upstream configuratio of pgAdmin 4 for PostgreSQL 11, specifically on a Flask dependency for compass.


cbandy (approved):
I like this config file much better.

PR #1173, opened 2019-12-19
[ch6662] Update container label tags

jkatz:
This matches the specified formats that are required by several
distributors, and removes the anti-pattern of having the LABEL subkeys
capitalized.

Issue: [ch6662]


crunchyjohn:
+1

PR #1174, opened 2019-12-20
Merge pgBackRest S3 Config & Wait for Cluster To Reach Consistent State

andrewlecuyer:
This commit ensures the `pgBackRest` settings required to enable both local and AWS S3 storage are merged with the existing `pgBackRest` configuration settings during cluster initialization, instead of overwriting the existing `pgBackRest` configuration settings.

Additionally, when an existing database is started manually in order to then be converted to a Patroni cluster, the container will now wait indefinitely for the cluster to reach a consistent state and become ready.  Additionally, if the container detects that PostgreSQL is no longer running while waiting for the DB to reach a consistent state, the container will attempt to start the database again (e.g. in the event that recovery failed due to unavailable WAL on the first start attempt due to a connectivity issue).  This will ensure the DB will keep attempting to recover without the container crashing, providing the opportunity to auto-recover and/or intervene and address any issues that may exist.

**Checklist:**

 
**Type of Changes:**

 
S3 configuration is applied instead of the base pgBackRest configuration.

The check used while waiting for the DB to reach a consistent state will timeout causing the container to exit.

S3 configuration is merged into the base pgBackRest configuration.

When an existing database is started manually in order to then be converted to a Patroni cluster, the container will now wait indefinitely for the cluster to reach a consistent state and become ready.

**Other information**:

N/A

### DIFF
--- a/bin/postgres-ha/bootstrap-postgres-ha.sh
+++ b/bin/postgres-ha/bootstrap-postgres-ha.sh
@@ -158,7 +158,18 @@ then
     echo_info "Database manually started"
 
     echo_info "Waiting to reach a consistent state"
-    test_server "postgres" "${PGHOST}" "${PGHA_PG_PORT}" "postgres"
+    until pg_isready --dbname="postgres" --host="${PGHOST}" --port="${PGHA_PG_PORT}" --username="postgres"
+    do
+        "Database has not reached a consistent state, sleeping..."
+        # sleep to give recovery a chance to complete
+        sleep 30
+        # if no postgres process is running at this point then assume a failed start and attempt to
+        # start again.  Otherwise check once again to see if recovery is complete
+        if ! pgrep postgres
+        then
+            pg_ctl -D "${PATRONI_POSTGRESQL_DATA_DIR}" -o "${manual_start_pg_ctl_options}" start
+        fi
+    done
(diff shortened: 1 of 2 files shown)


bin/postgres-ha/bootstrap-postgres-ha.sh
```diff
@@ -158,7 +158,18 @@ then
     echo_info "Database manually started"
 
     echo_info "Waiting to reach a consistent state"
-    test_server "postgres" "${PGHOST}" "${PGHA_PG_PORT}" "postgres"
+    until pg_isready --dbname="postgres" --host="${PGHOST}" --port="${PGHA_PG_PORT}" --username="postgres"
+    do
+        "Database has not reached a consistent state, sleeping..."
+        # sleep to give recovery a chance to complete
+        sleep 30
```
jkatz: @andrewlecuyer Though I think I had suggested this value yesterday, I wonder if this is too long, e.g. if this is the only DB that can be brought up in the area? Maybe we should go back to 5....even if it's a long time recovery. I think the cost is low.

andrewlecuyer: Sleep is now back to 5 seconds.

jkatz (approved):

PR #1178, opened 2019-12-28
Enable SSL in PG Database & Enable Client Cert for Replication User 

andrewlecuyer:
With this change, it is possible to apply the required SSL configuration to enable SSL in the PG database.  Specifically, by mounting the appropriate certificates in the `pgconf` directory, it is possible to  configure the `ssl_key_file`, `ssl_cert_file`,`ssl_ca_file` and `ssl_crl_file` `postgresql.conf` settings for the cluster via the DCS, effectively enabling the use of SSL and client certificates when authenticating into the database.

In support of configuring the DCS for SSL, the mechanism for detecting that the local node has been
initialized has been updated.  Now, in addition to verifying that the node is in a running state, a check is performed to first determine whether the local node has the primary or replica role, and then whether or not the primary or replica endpoint (depending on the role identified) is returning status code 200.  This ensures that all initial configuration has been applied to the node prior to updating the DCS with the SSL configuration using a JSON patch.

And finally, it is now possible to specify the certificates needed to allow the replication user to authenticate via client certificate in lieu of using a replication password.

**Checklist:**

 
**Type of Changes:**

 
- It is not possible to easily enable SSL for a cluster using the `crunchy-postgres-ha` container
- A password is required for the replication user

[ch6755]

- It is now possible to easily enable SSL for a cluster using the `crunchy-postgres-ha` container
- A client certificate can be used instead of a password for the replication user 

**Other information**:

N/A

### DIFF
--- a/bin/postgres-ha/bootstrap-postgres-ha.sh
+++ b/bin/postgres-ha/bootstrap-postgres-ha.sh
@@ -246,7 +246,8 @@ remove_patroni_pause_key
  # Ensure any existing SSL certificates in PGDATA have the proper permissions
 chmod -f 0600 "${PATRONI_POSTGRESQL_DATA_DIR}/server.key" "${PATRONI_POSTGRESQL_DATA_DIR}/server.crt" \
     "${PATRONI_POSTGRESQL_DATA_DIR}/ca.crt" "${PATRONI_POSTGRESQL_DATA_DIR}/ca.crl" \
-    "${PATRONI_POSTGRESQL_DATA_DIR}/replicator.crt" "${PATRONI_POSTGRESQL_DATA_DIR}/replicator.key"
+    "${PATRONI_POSTGRESQL_DATA_DIR}/replicator.crt" "${PATRONI_POSTGRESQL_DATA_DIR}/replicator.key" \
+    "${PATRONI_POSTGRESQL_DATA_DIR}/replicator.crl"
 
 # Bootstrap the cluster
 bootstrap_cmd="$@ /tmp/postgres-ha-bootstrap.yaml"
--- a/bin/postgres-ha/bootstrap-postgres-ha.sh
+++ b/bin/postgres-ha/bootstrap-postgres-ha.sh
@@ -43,6 +43,7 @@ trap_sigterm() {
 initialization_monitor() {
     echo_info "Starting background process to monitor Patroni initization and restart the database if needed"
     {
+        # Wait for the local node to enter a "running" state
         while [[ $(curl --silent "127.0.0.1:${PGHA_PATRONI_PORT}/patroni" --stderr - \
             | /opt/cpm/bin/yq r - state 2> /dev/null) != "running" ]]
         do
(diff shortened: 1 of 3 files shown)


bin/postgres-ha/bootstrap-postgres-ha.sh
```diff
@@ -43,15 +43,33 @@ trap_sigterm() {
...
+            sleep 1
+            echo "${init_role} not yet inititialized, retrying" >> "/tmp/patroni_initialize_check.log"
+
+            status_code=$(curl -o /dev/stderr -w "%{http_code}" "127.0.0.1:${PGHA_PATRONI_PORT}/${init_role}" 2> /dev/null)
+        done
+
+        # Apply custom configuration other than custom 'postgres-ha.yaml', e.g. custom keys and
+        # certificates to enable SSL
+        source /opt/cpm/bin/ssl-config.sh
+        echo_info "SSL config is: ${PGHA_SSL_CONFIG}"
```
jkatz, line 71: While the PostgreSQL GUCs read "TLS", (I'm hoping) at some point that it is all going to switch over to `tls_*` given that SSL is not even supported in PostgreSQL.

That said, I am torn over if we should call this `ssl` or `tls`.

bin/postgres-ha/pre-bootstrap.sh
```diff
@@ -220,17 +220,26 @@ set_pg_user_credentials() {
...
     then
         echo_info "Setting 'replicator' credentials using file system"
         
+        # Configure certificate-based authentication for replication if proper certs are available.
+        # Otherwise use a password
+        if [[ -f "/pgconf/replicator.key" ]] && [[ -f "/pgconf/replicator.crt" ]] && [[ -f "/pgconf/ca.crt" ]]
+        then
+            export PATRONI_REPLICATION_SSLKEY="${PATRONI_POSTGRESQL_DATA_DIR}/replicator.key"
+            export PATRONI_REPLICATION_SSLCERT="${PATRONI_POSTGRESQL_DATA_DIR}/replicator.crt"
+            export PATRONI_REPLICATION_SSLROOTCERT="${PATRONI_POSTGRESQL_DATA_DIR}/ca.crt"
```
jkatz: This block is missing a reference to the `ca.crl` file. This is optional, but we should include it in one of the variables if it exists.

bin/postgres-ha/pre-bootstrap.sh
```diff
@@ -204,7 +204,7 @@ set_pg_user_credentials() {
         env_check_err "PGHA_USER_PASSWORD"
     fi
 
-    if [[ -d "/pgconf/pguser" ]]
+    if [[ -d "/pgconf/pgsuper" ]]
```
jkatz, line 207: I would strongly advise that we spell out `superuser` -- let's make it so someone reviewing it has to think less.

jkatz (changes requested):
A few changes requested.

The ones that I would require:

- inclusion of the CRL file in the env var list (unless I misread that one)
- `pgsuper` => `pgsuperuser`

bin/postgres-ha/bootstrap-postgres-ha.sh
andrewlecuyer: I followed the PG docs here which seems to use SSL, as do the settings themselves (e.g. `ssl_ca_file`):

https://www.postgresql.org/docs/12/ssl-tcp.html

jkatz: @andrewlecuyer Yes, the GUCs are still `ssl_*`, but SSL is not used underneath the covers anymore. The `ssl_*` names are kept around for compatibility / ease-of-upgrading at this point.

My suggestion is around trying to be a bit more forward and call it "TLS" vs. "SSL" in our labeling/output, while still respecting the PostgreSQL GUCs.

bin/postgres-ha/pre-bootstrap.sh
andrewlecuyer: The CRL file is now included.

jkatz (approved):
Per discussion, the `pgsuperuser` will be addressed at a later date.

PR #1185, opened 2020-01-23
consolidates rhel7/ubi7 Dockerfile sets

wilybrace:
With the requirement to build ubi7 from an unsubscribed RHEL
machine lifted, these Dockerfile sets can be consolidated
with docker build-args to specify the base OS.

This change sets up the Makefile to point to the rhel7 file
set when building for UBI, while governing the inclusion tags
with the BASEOS build arg

**Checklist:**

 
 - [X] Have you updated or added documentation for the change, as applicable?
 - [X] Have you tested your changes on all related environments with successful results, as applicable?

**Type of Changes:**

 

crunchyheath (approved):
From my testing it all looks good.  I am able to build for both UBI and RHEL images on PG9.5-12.

PR #1216, opened 2020-03-20
Add LLVM support for PostgreSQL 11+ containers

jkatz:
This requires installing both libllvm and the package that enables
it for PostgreSQL versions after 11.

Per upsream PostgreSQL, JIT compilation, which uses libllvm, is
off by default in PostgreSQL 11, but enabled in PostgreSQL 12 and
beyond.

**Checklist:**

 
**Type of Changes:**

 
Add support for libllvm in the PostgreSQL containers, which in turn allows JIT compilation.


wilybrace (approved):
While this would certainly be cleaner by adding a conditional yum execution in between the bulk list yum command and the clean command, I'm going to suggest that approach is a "future improvement".

jkatz:
@wilybrace I look forward to your future contribution :wink:

PR #1220, opened 2020-03-25
Use the latest buildah

jkatz:
This ensures buildah is properly installed at the latest across
CentOS 7, 8 and RHEL7 environments.


jkatz:
Needs much more work.

PR #1221, opened 2020-04-01
Always promote during recovery

cbandy:
When `hot_standby` is disabled and no `recovery_target_action` is set, PostgreSQL will shut down at the end of recovery. We want PostgreSQL to continue running to clearly signal successful recovery.

We cannot delegate recovery to Patroni because it interprets a data directory restored by pgBackRest to be a PostgreSQL crash. To remedy the suspected crash, Patroni clears all recovery settings before starting PostgreSQL.

**Checklist:**

 
**Type of Changes:**

 
**What is the current behavior?**

When starting the container with a data directory restored by pgBackRest with no --target-action, PostgreSQL shuts down repeatedly.

**What is the new behavior?**

When starting the container with a data directory that needs to go through recovery, finish the recovery (regardless of --target-action) before starting Patroni.

### DIFF
--- a/bin/postgres-ha/bootstrap-postgres-ha.sh
+++ b/bin/postgres-ha/bootstrap-postgres-ha.sh
@@ -168,32 +168,33 @@ then
         > "${PATRONI_POSTGRESQL_DATA_DIR}/postgresql.conf"
     fi
 
-    manual_start_pg_ctl_options=""
-    # detect if this is a server that needs to entre recovery mode, which is
-    # what might happen after an instance is cloned.
-    # if this is the case, we will want to have a warm standby
+    manual_start_pg_ctl_options=()
     if [[ -f "${PATRONI_POSTGRESQL_DATA_DIR}/recovery.conf" || -f "${PATRONI_POSTGRESQL_DATA_DIR}/recovery.signal" ]]
     then
-      echo_info "Discovered presence of a recovery.conf or recovery.signal file"
-      echo_info "Setting hot_standby=\"off\" until PostgreSQL reaches a consistent state"
-      manual_start_pg_ctl_options="-c hot_standby=off"
+        # Prevent remote connections until after recovery has completed.
+        # PostgreSQL is restarted later, through Patroni, without these settings.
+        manual_start_pg_ctl_options+=("-c hot_standby='off'")
+        manual_start_pg_ctl_options+=("-c listen_addresses=''")
+        manual_start_pg_ctl_options+=("-c recovery_target_action='promote'")
     fi
 
     echo_info "Starting database manually prior to starting Patroni"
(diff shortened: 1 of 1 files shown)


bin/postgres-ha/bootstrap-postgres-ha.sh
```diff
@@ -168,32 +168,33 @@ then
         > "${PATRONI_POSTGRESQL_DATA_DIR}/postgresql.conf"
     fi
 
-    manual_start_pg_ctl_options=""
-    # detect if this is a server that needs to entre recovery mode, which is
-    # what might happen after an instance is cloned.
-    # if this is the case, we will want to have a warm standby
```
jkatz, line 174: Why is this comment removed?

bin/postgres-ha/bootstrap-postgres-ha.sh
```diff
@@ -168,32 +168,33 @@ then
...
+    manual_start_pg_ctl_options=()
     if [[ -f "${PATRONI_POSTGRESQL_DATA_DIR}/recovery.conf" || -f "${PATRONI_POSTGRESQL_DATA_DIR}/recovery.signal" ]]
     then
-      echo_info "Discovered presence of a recovery.conf or recovery.signal file"
-      echo_info "Setting hot_standby=\"off\" until PostgreSQL reaches a consistent state"
```
jkatz, line 178: Why remove these?

jkatz (commented):
Looks good, have a couple of questions.

bin/postgres-ha/bootstrap-postgres-ha.sh
cbandy: The second line isn't accurate. We're setting more than that one parameter now, and they stay in place until the restart around L75.

The first line should probably stay. When PostgreSQL starts it also logs that it is going into recovery, but if it _doesn't_ start this is the only indication we mucked with things.

bin/postgres-ha/bootstrap-postgres-ha.sh
cbandy: My understanding is that we control recovery because Patroni won't. That's probably what should be described here. We set some parameters to get the recovery behavior we want. "warm standby" encompasses recovery behavior but it wasn't clear to me. I intended for the new comment on L174 to capture this.

The concept of "cloning" seemed out of place at this layer, but that could just be me. I'll put some comments before the conditional. Is there anything you'd like to see there?

bin/postgres-ha/bootstrap-postgres-ha.sh
jkatz: OK.

bin/postgres-ha/bootstrap-postgres-ha.sh
jkatz: Along the lines of what you suggest. However, the `"warm standby"` bit is important, because you can recover as a "hot standby" which will cause some behavior changes in terms of detection (unless those were also modified)

cbandy:
Superseded by #1222.

PR #1224, opened 2020-04-20
Fix WAL directory configuration

cbandy:
**Checklist:**

 
**Type of Changes:**

 
`postgres-ha-waldir.yaml` sets one `initdb` argument, but `postgres-ha-initdb.yaml` overwrites it later.

When `PGHA_WALDIR` is non-empty,
- it is appended to `initdb` arguments after `postgres-ha-initdb.yaml` is applied.
- its effect is preserved during `pgbackrest restore`.
- it works for PostgreSQL 9.6 and earlier.

### DIFF
--- a/bin/postgres-ha/bootstrap/pre-bootstrap.sh
+++ b/bin/postgres-ha/bootstrap/pre-bootstrap.sh
@@ -40,12 +40,6 @@ set_default_pgha_autoconfig_env()  {
         default_pgha_autoconfig_env_vars+=("PGHA_BASE_PG_CONFIG")
     fi
 
-    if [[ ! -v PGHA_ENABLE_WALDIR ]]
-    then
-        export PGHA_ENABLE_WALDIR="false"
-        default_pgha_autoconfig_env_vars+=("PGHA_ENABLE_WALDIR")
-    fi
-
     if [[ ! -v PGHA_PGBACKREST ]]
     then
         export PGHA_PGBACKREST="true"
@@ -109,18 +103,6 @@ set_default_pgha_env()  {
     then
         export PGHA_DATABASE="userdb"
         default_pgha_env_vars+=("PGHA_DATABASE")
-    fi
-
-    if [[ "${PGHA_ENABLE_WALDIR}" == "true" ]]  # set PGHA_ENABLE_WALDIR to "false" if not "true"
-    then
-        if [[ ! -v PGHA_WALDIR ]]
-        then
-            export PGHA_WALDIR="/pgwal/${HOSTNAME}-wal"
-            default_pgha_env_vars+=("PGHA_WALDIR")
-        fi
-    else
-        echo_info "The use of the /pgwal directory for writing WAL is not enabled"
-        echo_info "A default value will not be set for PGHA_WALDIR and any value provided for will be ignored"
     fi
 
     if [[ ! -v PGHA_REPLICA_REINIT_ON_START_FAIL ]]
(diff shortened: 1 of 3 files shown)


bin/postgres-ha/bootstrap/pre-bootstrap.sh
```diff
@@ -322,6 +293,12 @@ build_bootstrap_config_file() {
         /opt/cpm/bin/yq m -i -x "${bootstrap_file}" "/opt/cpm/conf/postgres-ha-initdb.yaml"
     fi
 
+    if [[ -n "${PGHA_WALDIR}" ]]
+    then
+        echo_info "Applying custom WAL dir to postgres-ha configuration"
+        /opt/cpm/bin/yq w -i "${bootstrap_file}" 'bootstrap.initdb[+].waldir' "${PGHA_WALDIR}"
```
cbandy: This only has an effect when `initdb` is called. I wonder if it should move into the `"${PGHA_INIT}" == "true"` block above to make that more obvious?

abrightwell: It would seem logical, IMO... :thinking:  :man_shrugging:

jkatz: @cbandy Can it only ever be called once and on init? Seemingly yes. Probably probably be moved then, of course with proper testing.

cbandy:
Hold on review. I've got some changes coming.

jkatz (approved):

bin/postgres-ha/bootstrap/pre-bootstrap.sh
andrewlecuyer: This made me realize are prob some other configs that can now be left exclusively for the init config as well - essentially anything written to the `dcs` and/ or `initdb` sections. `initdb` is obviously only applicable at cluster creation, and all replicas inheriting DCS config from the actual DCS `configMap` once the cluster has been bootstrapped, so really only local config needs to be included when creating a new replica.

Very early on when designed as a standalone container, `crunchy-postgres-ha` was actually without the concept of an "init" node.  Basically, more than one node could be created at the same time when creating a new cluster, and all will fight to become leader and bootstrap the cluster.  This therefore meant that all those nodes would need the same init and DCS configs, in case they ultimately became responsible for bootstrapping the cluster.

However, as this container continues to evolve into what is essentially a PGO-exclusive container, (and since the Operator requires additional orchestration), these "standalone" container ideas and concepts are really no longer applicable, so we can rethink some of this design.  One thing that also comes to mind here are some of the places we set the `initdb` setting back to `true` (e.g. post-restore or for a cloned cluster), which would result in the execution of this logic, even though the `initdb` stuff isn't needed since the DB already exists (having it there is harmless, just something to consider cleaning up moving forward).

So anyway, just some food for thought moving forward since this got me thinking a bit.  For now I definitely think it makes sense to only include the WAL config when `PHGA_INIT` is `true`.

bin/postgres-ha/bootstrap/pre-bootstrap.sh
```diff
@@ -322,6 +293,17 @@ build_bootstrap_config_file() {
...
     fi
 
+    if [[ -n "${PGHA_WALDIR}" ]]
+    then
+        echo_info "Applying custom WAL dir to postgres-ha configuration"
+        if printf '10\n'${PGVERSION} | sort -VC
+        then
+            /opt/cpm/bin/yq w -i "${bootstrap_file}" 'bootstrap.initdb[+].waldir' "${PGHA_WALDIR}"
+        else
+            /opt/cpm/bin/yq w -i "${bootstrap_file}" 'bootstrap.initdb[+].xlogdir' "${PGHA_WALDIR}"
```
andrewlecuyer: Just a note that I'll be bumping up the version of `yq` to the latest version (`v3.2.1`) with a PR today, so just want to mention since I'm trying to ensure all `yq` functionality continues to work as expected.  Def like the approach here and eliminating a `yaml` file.

andrewlecuyer (approved):
Looks good.

cbandy:
Rebased and tested (with yq 3.3.0). Moved WAL code into init conditional.

jkatz (approved):
please merge when ready.

PR #1227, opened 2020-04-22
Consolidate pg_upgrade Dockerfiles

jkatz:
**Checklist:**

 
**Type of Changes:**

 
This commit merges the `pg_upgrade` CentOS/RHEL/UBI Dockerfiles into a single file
and leverages the ability to utilize multiple repository files with
wildcards.

This also removes some of the special Makefile targets used by the
upgrade container as they can now leverage the standard ones.


centos7/Dockerfile.upgrade.centos7
```diff
@@ -1,37 +1,45 @@
 ARG BASEOS
 ARG BASEVER
 ARG PG_FULL
+ARG PG_MAJOR
 ARG PREFIX
 FROM ${PREFIX}/crunchy-pg-base:${BASEOS}-${PG_FULL}-${BASEVER}
```
cbandy: Does this work when building with Docker? My understanding is that any `ARG` before `FROM` only applies to `FROM`: https://docs.docker.com/engine/reference/builder/#understand-how-arg-and-from-interact

If we're relying on buildah being different, that may have changed recently: [containers/buildah#2220](https://github.com/containers/buildah/pull/2220)

centos7/Dockerfile.upgrade.centos7
```diff
@@ -1,37 +1,45 @@
...
+# add in all of the earlier version of PostgreSQL. It will install the version
+# above, but the dependencies are handled
+RUN yum -y install \
+	--setopt=skip_missing_names_on_install=False \
+	--disablerepo=* \
+	--enablerepo=crunchypg* \
+	postgresql[1-9][0-9] \
+	postgresql[1-9][1-9]-contrib \
+	postgresql[1-9][1-9]-server \
+	pgaudit[1-9][1-9] \
```
cbandy: Should all these number ranges match? Why does the first one have a zero but the others don't?

centos7/Dockerfile.upgrade.centos7
jkatz: I'm fine with we swap the from. That said, we're only building with `buildah`

centos7/Dockerfile.upgrade.centos7
jkatz: Oh I messed up, the second ones should be `[0-9]` :|

PR #1231, opened 2020-04-29
Upgrade + PostgreSQL Base Image Fixes

jkatz:
## Ensure upgrade consolidation works for PostgreSQL 9.6

The build argument PG_MAJOR container a "." in it, which did not
mesh correctly with the repo file. This subs out the "."

## Set proper repository enablement in PostgreSQL base container

This ensures that only the appropriate base PostgreSQL files are
installed in a given environment.


crunchyheath (approved):
+1 from me.  I think this is a good first step.  To completely safeguard users from pulling in content from the wrong location we would have to add these disable/enables changes to any Dockerfile that pulls from our repos.

PR #1242, opened 2020-06-24
Enable replication connections to use certificate-based authentication

jkatz:
**Checklist:**

 
**Type of Changes:**

 
[Certificate-based authentication](https://www.postgresql.org/docs/current/auth-cert.html) is a powerful PostgreSQL feature
that allows for a PostgreSQL client to authenticate using a TLS
certificate. While there are a variety of permutations for this can be
set up, we can at least create a standardized way for enabling the
replication connection to authenticate with a certificate, as we do have
a known certificate authority.

This introduces support for certificate-based authentication for the
replication user, presuming that the CN on the certificate is that of
`PATRONI_REPLICATION_USERNAME` (default is "primaryuser"). If the CN
on the certificate is different, one must include a "pg_ident"
customization as part of their configuration.

If TLS is enabled on the container and a Secret that contains the
replication user's TLS credentials is mounted to the container, then
certificate-based authentication for this user is automatically set up.
As the CA must be appropriately mounted for this to work, we are able
to set PGSSLMODE to use "verify-ca" by default, to include a stronger
guarnatee of validity between the instances (though not as strong as
"verify-full", but we cannot make that level of assumption about the
environment).

Issue: [ch8387]

### DIFF
--- a/bin/postgres-ha/bootstrap/pre-bootstrap.sh
+++ b/bin/postgres-ha/bootstrap/pre-bootstrap.sh
@@ -318,8 +318,47 @@ build_bootstrap_config_file() {
     # postgresql.conf file
     if [[ "${PGHA_TLS_ENABLED}" == "true" ]]
     then
+      echo_info "Enabling TLS"
+      pghba_tls_file="/opt/cpm/conf/postgres-ha-pghba-tls.yaml"
+
+      # if there is a TLS keypair for a replication user detected, we will need
+      # to copy this to a special mount in order to properly enable
+      # certificate-based authentication between replicas. This is due to
+      # PostgreSQL requiring a client's TLS key to have permissions no less strict
+      # than 0600. In this case, the client is the "postgres" user, and because
+      # Kubernetes Secrets are owned by the group, we need to create a copy of
+      # this key with the correct permissions
+      if [[ -f "/pgconf/tls/tls-replication.key" ]] && [[ -f "/pgconf/tls/tls-replication.crt" ]]
+      then
+        echo_info "Enabling certificate-based authentication for replication"
+
+        tls_replication_key_file="/pgconf/tls-replication/tls.key"
+        tls_replication_cert_file="/pgconf/tls-replication/tls.crt"
+        touch "${tls_replication_key_file}" "${tls_replication_cert_file}"
+        install -m 0600 /pgconf/tls/tls-replication.key "${tls_replication_key_file}"
+        install -m 0600 /pgconf/tls/tls-replication.crt "${tls_replication_cert_file}"
+
+        # update the bootstrap parameters to set the certificate-based
+        # authentication settings
+        /opt/cpm/bin/yq w -i "${bootstrap_file}" postgresql.authentication.replication.sslkey "${tls_replication_key_file}"
+        /opt/cpm/bin/yq w -i "${bootstrap_file}" postgresql.authentication.replication.sslcert "${tls_replication_cert_file}"
+        /opt/cpm/bin/yq w -i "${bootstrap_file}" postgresql.authentication.replication.sslrootcert "/pgconf/tls/ca.crt"
+        # as we have the CA available, we can make the TLS mode "verify-ca" by
+        # default
+        /opt/cpm/bin/yq w -i "${bootstrap_file}" postgresql.authentication.replication.sslmode "verify-ca"
+        # if a CRL file is provided, add this to the stanza as well
+        if [[ -f "/pgconf/tls/ca.crl" ]]
+        then
+          /opt/cpm/bin/yq w -i "${bootstrap_file}" postgresql.authentication.replication.sslcrl "/pgconf/tls/ca.crl"
+        fi
+
+        # use an updated pg_hba.conf file that enforces certificate based
+        # authentication
+        pghba_tls_file="/opt/cpm/conf/postgres-ha-pghba-tls-auth.yaml"
+      fi
+
       echo_info "Applying TLS remote connection configuration to pg_hba.conf"
-      /opt/cpm/bin/yq m -i -a "${pghba_file}" "/opt/cpm/conf/postgres-ha-pghba-tls.yaml"
+      /opt/cpm/bin/yq m -i -a "${pghba_file}" "${pghba_tls_file}"
       echo_info "Enabling TLS in postgresql.conf"
       /opt/cpm/bin/yq m -i -a "${bootstrap_file}" "/opt/cpm/conf/postgres-ha-pgconf-tls.yaml"
 
--- a/conf/postgres-ha/postgres-ha-pghba-tls-auth.yaml
+++ b/conf/postgres-ha/postgres-ha-pghba-tls-auth.yaml
@@ -1,0 +1,6 @@
+---
+postgresql:
+  pg_hba:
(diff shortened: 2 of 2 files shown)


andrewlecuyer (changes requested):
Came across a few things I wanted to run by you.

bin/postgres-ha/bootstrap/pre-bootstrap.sh
```diff
@@ -318,8 +318,46 @@ build_bootstrap_config_file() {
...
+      # to copy this to a special mount in order to properly enable
+      # certificate-based authentication between replicas. This is due to
+      # PostgreSQL requiring a client's TLS key to have permissions no less strict
+      # than 0600. In this case, the client is the "postgres" user, and because
+      # Kubernetes Secrets are owned by the group, we need to create a copy of
+      # this key with the correct permissions
+      if [[ -f "/pgconf/tls/tls-replication.key" ]] && [[ -f "/pgconf/tls/tls-replication.crt" ]]
+      then
+        echo_info "Enabling certificate-based authentication for replication"
+        touch /pgconf/tls-replication/tls.key /pgconf/tls-replication/tls.crt
```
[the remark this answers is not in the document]
jkatz: @andrewlecuyer `/pgconf/tls-replication` is an `emptyDir`, see: [https://github.com/CrunchyData/postgres-operator/pull/1662/files#diff-f3ecf8804e14f07041039de1d908c047](https://github.com/CrunchyData/postgres-operator/pull/1662/files#diff-f3ecf8804e14f07041039de1d908c047)

bin/postgres-ha/bootstrap/pre-bootstrap.sh
```diff
@@ -318,8 +318,46 @@ build_bootstrap_config_file() {
...
+        /opt/cpm/bin/yq w -i "${bootstrap_file}" postgresql.authentication.replication.sslmode "verify-ca"
+        # if a CRL file is provided, add this to the stanza as well
+        if [[ -f "/pgconf/tls/ca.crl" ]]
+        then
+          /opt/cpm/bin/yq w -i "${bootstrap_file}" postgresql.authentication.replication.sslcrl "/pgconf/tls/ca.crl"
+        fi
+
+        # use an updated pg_hba.conf file that enforces certificate based
+        # authentication
+        pghba_tls_file="/opt/cpm/conf/postgres-ha-pghba-tls-auth.yaml"
```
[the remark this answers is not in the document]
jkatz: I may have forgotten to commit that!

bin/postgres-ha/bootstrap/pre-bootstrap.sh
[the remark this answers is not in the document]
andrewlecuyer: Thanks for the clarification.  I assume the fact that other ConfigMaps and Secrets are using the `pgconf` directory doesn't impact the the writeability of the `emptyDir` under `pgconf`?

Also, would it makes sense to just use another existing directory on the file system, e.g. `tmp` or `crunchyadm`, instead of mounting another `emptyDir`?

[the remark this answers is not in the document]
jkatz: @andrewlecuyer AFAICT it does not impact the writability, given I've tested this and it works in my Kubernetes environment. I will admit to it needing testing in OpenShift.

It feels grotty to me to put TLS credentials into `/tmp` and it would make me wonder what happens if someone wants to run a readOnly root FS (see [ch7571] -- and technically we can't support this fully yet).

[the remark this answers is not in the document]
andrewlecuyer: Doesn't need to be `tmp`, was just thinking that or something similar (e.g. we have our `crunchyadm` directory that isn't used much).  I know there have been questions about our various `emptyDirs` in the past, so figured I'd ask (I also tend to think of standalone containers and portability, but that doesn't apply so much with this container).

Definitely functionally the same, so I'm good either way.  Thanks!

andrewlecuyer (approved):

bin/postgres-ha/bootstrap/pre-bootstrap.sh
```diff
@@ -318,8 +318,46 @@ build_bootstrap_config_file() {
...
+      # certificate-based authentication between replicas. This is due to
+      # PostgreSQL requiring a client's TLS key to have permissions no less strict
+      # than 0600. In this case, the client is the "postgres" user, and because
+      # Kubernetes Secrets are owned by the group, we need to create a copy of
+      # this key with the correct permissions
+      if [[ -f "/pgconf/tls/tls-replication.key" ]] && [[ -f "/pgconf/tls/tls-replication.crt" ]]
+      then
+        echo_info "Enabling certificate-based authentication for replication"
+        touch /pgconf/tls-replication/tls.key /pgconf/tls-replication/tls.crt
+        chown postgres:postgres /pgconf/tls-replication/tls.key /pgconf/tls-replication/tls.crt
```
cbandy: Is this necessary? I would have expected `touch` to create a file owned by the caller.

bin/postgres-ha/bootstrap/pre-bootstrap.sh
```diff
@@ -318,8 +318,46 @@ build_bootstrap_config_file() {
...
+        chmod 0600 /pgconf/tls-replication/tls.key /pgconf/tls-replication/tls.crt
+        cat /pgconf/tls/tls-replication.key > /pgconf/tls-replication/tls.key
+        cat /pgconf/tls/tls-replication.crt > /pgconf/tls-replication/tls.crt
```
cbandy: Can all this be accomplished with `install`?

```suggestion
        install -m 0600 /pgconf/tls/tls-replication.key /pgconf/tls-replication/tls.key
        install -m 0600 /pgconf/tls/tls-replication.crt /pgconf/tls-replication/tls.crt
```

cbandy: I expected the origin to be shaped like a Kubernetes TLS Secret (one directory containing `tls.key` and `tls.crt`).

Is the destination shaped this way in case the Secret can be mounted directly in the future?

jkatz: it is. it's included as part of a projected volume that has another TLS secret. You need to view the corresponding patch at CrunchyData/postgres-operator#1662

cbandy: Bleh. `install` sets permissions _after_ the contents have been copied.

jkatz: That's OK. I verified and this worked.

bin/postgres-ha/bootstrap/pre-bootstrap.sh
jkatz: Nope, that was done out of an abundance of paranoia. Removed.

bin/postgres-ha/bootstrap/pre-bootstrap.sh
jkatz: Leaving this as is.

bin/postgres-ha/bootstrap/pre-bootstrap.sh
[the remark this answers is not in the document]
jkatz: Done.

PR #1246, opened 2020-07-01
Escape password values stored in PostgreSQL password file

tjmoore4:
**Checklist:**

 
**Type of Changes:**

 
Currently, PostgreSQL passwords stored in the .pgpass file
are not properly escaped.
As documented here:
https://www.postgresql.org/docs/current/libpq-pgpass.html
"If an entry needs to contain : or \, escape this character with \"

This update adds that functionality to the passwords stored in
the .pgpass files used by the pgdump, pgbench, pgbackup and
pgrestore container scripts.

**Other information**:

### DIFF
--- a/bin/backup/start-backupjob.sh
+++ b/bin/backup/start-backupjob.sh
@@ -40,7 +40,11 @@ env_check_info "BACKUP_LABEL" "BACKUP_LABEL is set to ${BACKUP_LABEL}."
 env_check_info "BACKUP_PATH" "BACKUP_PATH is set to ${BACKUP_PATH}."
 env_check_info "BACKUP_OPTS" "BACKUP_OPTS is set to ${BACKUP_OPTS}."
 
-echo "*:*:*:"${BACKUP_USER?}":"${BACKUP_PASS?} >> ${PGPASSFILE?}
+# escape any instances of ':' or '\' with '\' in the provided password
+# before storing the value in the password file
+ESCAPED_PASSWORD=$(sed <<< "${BACKUP_PASS?}" 's/[:\\]/\\&/g')
+
+echo "*:*:*:${BACKUP_USER?}:${ESCAPED_PASSWORD?}" >> ${PGPASSFILE?}
 chmod 600 ${PGPASSFILE?}
 
 pg_basebackup --label=${BACKUP_LABEL?} -X fetch \
--- a/bin/pgbench/start.sh
+++ b/bin/pgbench/start.sh
@@ -42,8 +42,13 @@ fi
 
 function create_pgpass() {
     cd /tmp
+
+# escape any instances of ':' or '\' with '\' in the provided password
+# before storing the value in the password file
+ESCAPED_PASSWORD=$(sed <<< "${PG_PASSWORD?}" 's/[:\\]/\\&/g')
+
 cat >> ".pgpass" <<-EOF
-*:*:*:${PG_USERNAME?}:${PG_PASSWORD?}
+*:*:*:${PG_USERNAME?}:${ESCAPED_PASSWORD?}
 EOF
     chmod 0600 .pgpass
 }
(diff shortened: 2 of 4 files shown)


jkatz (approved):
LGTM

bin/backup/start-backupjob.sh
```diff
@@ -40,7 +40,11 @@ env_check_info "BACKUP_LABEL" "BACKUP_LABEL is set to ${BACKUP_LABEL}."
 env_check_info "BACKUP_PATH" "BACKUP_PATH is set to ${BACKUP_PATH}."
 env_check_info "BACKUP_OPTS" "BACKUP_OPTS is set to ${BACKUP_OPTS}."
 
-echo "*:*:*:"${BACKUP_USER?}":"${BACKUP_PASS?} >> ${PGPASSFILE?}
+# escape any instances of ':' or '\' with '\' in the provided password
+# before storing the value in the password file
+ESCAPED_PASSWORD=$(echo ${BACKUP_PASS?} | sed 's/[:\\]/\\&/g')
```
cbandy: I like the simplicity of this `sed` expression. We can eliminate the `echo` using [Bash here string](https://wiki.bash-hackers.org/syntax/redirection#here_strings). With this, the assignment will have the same exit code as `sed`.

```suggestion
ESCAPED_PASSWORD=$(sed <<< "${BACKUP_PASS?}" 's/[:\\]/\\&/g')
```

---

If we _don't_ want to use `sed`, I think the same effect can be had with Bash parameter expansion. It's easier to get wrong (backslash needs to be replaced before colon; don't accidentally swap them) but it does save one process spawn. I prefer `sed`, I think.

```suggestion
ESCAPED_PASSWORD="${BACKUP_PASS?}"
ESCAPED_PASSWORD="${ESCAPED_PASSWORD//\\/\\\\}"
ESCAPED_PASSWORD="${ESCAPED_PASSWORD//:/\\:}"
```

tjmoore4: I like the here string change, I'll implement that. I think, with sed being a bit easier to 'get right', I'll also agree to stick with that. Thanks!

bin/backup/start-backupjob.sh
```diff
@@ -40,7 +40,11 @@ env_check_info "BACKUP_LABEL" "BACKUP_LABEL is set to ${BACKUP_LABEL}."
 env_check_info "BACKUP_PATH" "BACKUP_PATH is set to ${BACKUP_PATH}."
 env_check_info "BACKUP_OPTS" "BACKUP_OPTS is set to ${BACKUP_OPTS}."
 
-echo "*:*:*:"${BACKUP_USER?}":"${BACKUP_PASS?} >> ${PGPASSFILE?}
+# escape any instances of ':' or '\' with '\' in the provided password
+# before storing the value in the password file
+ESCAPED_PASSWORD=$(sed <<< "${BACKUP_PASS?}" 's/[:\\]/\\&/g')
+
+echo "*:*:*:"${BACKUP_USER?}":"${ESCAPED_PASSWORD?} >> ${PGPASSFILE?}
```
cbandy: I didn't notice that these expansions are outside quotes. As-is, this will mess up values with consecutive whitespace. Instead,

```suggestion
echo "*:*:*:${BACKUP_USER?}:${ESCAPED_PASSWORD?}" >> ${PGPASSFILE?}
```

```
BACKUP_USER="bob"
ESCAPED_PASSWORD="a  b  c"

echo "*:*:*:"${BACKUP_USER?}":"${ESCAPED_PASSWORD?} | hexdump -C
00000000  2a 3a 2a 3a 2a 3a 62 6f  62 3a 61 20 62 20 63 0a  |*:*:*:bob:a b c.|
00000010

echo "*:*:*:${BACKUP_USER?}:${ESCAPED_PASSWORD?}" | hexdump -C
00000000  2a 3a 2a 3a 2a 3a 62 6f  62 3a 61 20 20 62 20 20  |*:*:*:bob:a  b  |
00000010  63 0a                                             |c.|
00000012
```

(The other scripts aren't affected because they use a heredoc.)

cbandy (changes requested):
Noticed one more issue, then LGTM.

bin/backup/start-backupjob.sh
tjmoore4: gotta, good catch, thanks.

cbandy (approved):

andrewlecuyer (approved):

PR #1247, opened 2020-07-08
Ensure pgBackRest Settings Work When Local & S3 Storage is Enabled

tjmoore4:
**Checklist:**

 
**Type of Changes:**

 
Currently, pgBackRest backups initiated by the Postgres Operator
that utilize the 'local,s3' setting do not work as expected with the 
's3-verify-tls' & 's3-uri-style' configuration settings.

This update is the companion update to the Postgres Operator PR which
ensures that the 's3-verify-tls' & 's3-uri-style' settings work as expected
when both local & S3 storage are enabled for pgBackRest when creating a
pgcluster. Currently, when a new cluster is created using Local and AWS S3
storage (e.g. by specifiying --pgbackrest-storage-type=local,s3 with the
pgo create command), stanza creation fails and the cluster is not initialized.

Since this is failing because the 'non S3' pgBackRest command runs when
the repo1-s3-verify-tls flag is detected in the environment. This fix updates
the pgBackRest flag implementation so that the repo1-s3-verify-tls setting
is applied only for the second pgbackrest command only, i.e. the command using
--repo-type-s3, and not apply it when running the first pgbackrest command for
the local repository. This is done by using the --no-repo1-s3-verify-tls
command flag rather than the relevant environment variable or the
related configuration file setting. This method allows both commands to
execute as expected.

**Other information**:
[ch8634]

### DIFF
--- a/bin/postgres-ha/pgbackrest/pgbackrest-archive-push-local-s3.sh
+++ b/bin/postgres-ha/pgbackrest/pgbackrest-archive-push-local-s3.sh
@@ -30,8 +30,18 @@ source /opt/cpm/bin/pgbackrest/pgbackrest-set-env.sh
 pgbackrest archive-push $1
 local_exit=$?
 
+# set the repo type flag
+archive_push_cmd_args=("--repo1-type=s3")
+
+# if TLS verification is disabled, pass in the appropriate flag
+# otherwise, leave the default behavior and verify TLS
+if [[ $PGHA_PGBACKREST_S3_VERIFY_TLS == "false" ]]
+then
+    archive_push_cmd_args+=("--no-repo1-s3-verify-tls")
+fi
+
 # then try S3
-pgbackrest archive-push --repo-type=s3 $1
+pgbackrest archive-push ${archive_push_cmd_args[*]} $1
 s3_exit=$?
 
 # check each exit code. If one of them fail, exit with their nonzero exit code
--- a/bin/postgres-ha/pgbackrest/pgbackrest-create-replica.sh
+++ b/bin/postgres-ha/pgbackrest/pgbackrest-create-replica.sh
@@ -19,6 +19,8 @@ enable_debugging
 source /opt/cpm/bin/common/pgha-common.sh
 export $(get_patroni_pgdata_dir)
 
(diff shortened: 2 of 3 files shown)


bin/postgres-ha/pgbackrest/pgbackrest-archive-push-local-s3.sh
```diff
@@ -31,8 +31,17 @@ pgbackrest archive-push $1
...
-pgbackrest archive-push --repo-type=s3 $1
-s3_exit=$?
+# if TLS verification is disabled, pass in the appropriate flag
+# otherwise, leave the default behavior and verify TLS
+if [[ $PGHA_PGBACKREST_S3_VERIFY_TLS == "false" ]]
+then
+    pgbackrest archive-push --repo1-type=s3 --no-repo1-s3-verify-tls $1
+    s3_exit=$?
+elif
+then
```
jkatz: `else`?

jkatz: Hm looking at this, perhaps we set up something like:

```bash
archive_push_cmd_args=("--repo1-type=s3")

# if TLS verification is disabled, pass in the appropriate flag
# otherwise, leave the default behavior and verify TLS
if [[ $PGHA_PGBACKREST_S3_VERIFY_TLS == "false" ]]
then
    archive_push_cmd_args+=("--no-repo1-s3-verify-tls")
fi

pgbackrest archive-push ${archive_push_cmd_args[*]} $1
s3_exit=$?
```

tjmoore4: Yes, that should be an else, fixed, thanks.

And the update to the flag handling looks good to me, updated.

jkatz (commented):
@tjmoore4 This looks good to me. IS it still in draft mode?

tjmoore4:
@jkatz yes, at the moment. Just pushed a few additional tweaks, and we're working on some edge case testing. I'm planning to take this PR out of draft with its 'sister' PR: https://github.com/CrunchyData/postgres-operator/pull/1691

jkatz (approved):

PR #1261, opened 2020-08-06
Unification of Dockerfiles into a single Dockerfile per container

crunchyheath:
**Checklist:**

 
**Type of Changes:**

 
Currently there is a centos7 and a rhel7 Dockerfile for every single container.

This combines the 2 OS Dockerfiles into a single Dockerfile and allows for adding of additional support to the single Dockerfile.

**Other information**:
  I have done spot builds to verify that this will still build on UBI7 and CentOS7 OS's.  I have only done a handful of these builds but will continue to test and PR any breakages that happen to occur.


build/base/Dockerfile
```diff
@@ -0,0 +1,102 @@
...
+ARG DFSET
+ARG PACKAGER
+
+MAINTAINER Crunchy Data <info@crunchydata.com>
+
+LABEL vendor="Crunchy Data" \
+	url="https://crunchydata.com" \
+	release="${RELVER}" \
+	org.opencontainers.image.vendor="Crunchy Data" \
+	os.version="7.7"
```
cbandy, line 17: This `os.version` label doesn't seem relevant anymore.

build/pgbadger/Dockerfile
```diff
@@ -20,11 +23,37 @@ LABEL name="pgbadger" \
...
-		--enablerepo="rhel-7-server-optional-rpms" \
-		--setopt=skip_missing_names_on_install=False \
-		pgbadger \
-	&& yum -y clean all --enablerepo="rhel-7-server-optional-rpms"
+RUN if [ "$DFSET" = "centos" ] ; then \
+       ${PACKAGER} -y install \
+                --enablerepo="epel" \
+                --setopt=skip_missing_names_on_install=False \
+                pgbadger \
+        && yum -y clean all ; \
```
cbandy: `yum` is called here and a few more times in this file. Should it be `${PACKAGER}`?

Makefile
```diff
@@ -10,22 +10,56 @@ CCP_PATRONI_VERSION ?= 1.6.5
 CCP_BACKREST_VERSION ?= 2.27
 CCP_VERSION ?= 4.4.0
 CCP_POSTGIS_VERSION ?= 3.0
+CCP_BASEREGISTRY ?=
```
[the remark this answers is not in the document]
crunchyheath: Yeah I will remove that.  It was the first name of the argument that ended up being DOCKERBASEREGISTRY.  Thanks

build/base/Dockerfile
crunchyheath: That is possible, but that is not something I was going to change from what it is in the original Dockerfile.  We can have another PR for stripping those out if we need to once we verify that the scanners aren't looking for it.

build/pgbadger/Dockerfile
crunchyheath: Yes it should.  I will fix it.

crunchyheath:
I believe this last commit resolves the open questions.  Please let me know if you find any other issues.  Thank you.

Makefile
```diff
@@ -10,22 +10,55 @@ CCP_PATRONI_VERSION ?= 1.6.5
...
+ifeq ("$(CCP_BASEOS)", "rhel7")
+        DFSET=rhel
+endif
+
 ifeq ("$(CCP_BASEOS)", "ubi7")
-	DFSET=rhel7
+        DFSET=rhel
+endif
+
+ifeq ("$(CCP_BASEOS)", "rhel8")
```
jkatz: Do we need to support this?

build/base/Dockerfile
jkatz: I think better would be to add it to a build var so we can toggle it. OK-ish if it's a separate commit.

jkatz (commented):
Overall looks
Good - thanks for the efforts!

Left a couple of questions + comments but this is likely close, if not ready.

Makefile
crunchyheath: I was not sure if we would have to support it or not.  I did not add any of the surrounding code in the Dockerfiles to do so, but left it in just in case.  If you want me to remove it just let me know

jkatz: Yeah, let's remove it. If we don' thave it any any of the Dockerfiles, I don't see why we need it in the Makefile :smile:

jkatz (approved):

andrewlecuyer:
@crunchyheath looks like there are some conflicts preventing me from merging this PR.

Can you rebase against the current master, and I will then go ahead and merge?

PR #1278, opened 2020-09-17
Add wal2json to the PostgreSQL containers

### LINKED ISSUE #1816: Using wal2json in pgo-cluster
Hi,
I'm using PGO v4.4.0
I want to use plugin wal2json for logical decoding on my pgo-cluster.
But I am getting error:
bash-4.2$ pg_recvlogical -d postgres --slot test_slot --create-slot -P wal2json
pg_recvlogical: error: could not send replication command "CREATE_REPLICATION_SLOT "test_slot" LOGICAL "wal2json" NOEXPORT_SNAPSHOT": ERROR:  could not access file "wal2json": No such file or directory

How can I install the wal2json plugin on my pgo-cluster?

### LINKED ISSUE #1841: How can I run the pgo operator with a custom image?
I have replaced the values ​​ccp_image_prefix and pgo_image_prefix in the helm chart
to your custom image, but I ran into the fact that when replacing ccp_image_prefix --- pgo-deploy does not start, when replacing pgo_image_prefix: ----- tries to download the image from registry.developers.crunchydata.com/crunchydata/custom_container:tag

Сould you help with this?

jkatz:
wal2json is great for applications that want to leverage
logical decoding in an output format that is easy to understand,
let alone parse. The committer of the patch has even used it
himself in the past. So, let's bring it into the containers.

Issue: [ch9125]
fixes CrunchyData/postgres-operator#1816
fixes CrunchyData/postgres-operator#1841

### DIFF
--- a/build/postgres-ha/Dockerfile
+++ b/build/postgres-ha/Dockerfile
@@ -34,6 +34,7 @@ RUN if [ "$DFSET" = "centos" ] ; then \
         	psmisc \
         	rsync \
         	less \
+        	wal2json${PG_MAJOR//.} \
         && ${PACKAGER} -y clean all ; \
 else \
        ${PACKAGER} -y install \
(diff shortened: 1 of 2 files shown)


crunchyjohn:
Seems like it would work.  +1

PR #1282, opened 2020-09-22
Modify PL/Python install directive

jkatz:
As the PostgreSQL 13 builds drop support for Python 2, we need
to modify the install directive so we can include Python 2 on
older installs as well as Python 3 everywhere.


crunchyheath (approved):
+1 from me.

PR #1306, opened 2020-12-15
Upgrade compaction

jmckulk:
This change updates the crunchy-upgrade container to be based off of
crunchy-base image instead of crunchy-pg-base. This is one step closer to
removing the crunchy-pg-base image entirely. The binary and config paths are
also updated to use `/opt/crunchy` instead of `/opt/cpm`.

This set of changes also includes two bug fixes for the upgrade container:

First, the pgnodemx package has been added to the crunchy-postgres image and is now
included in the default postgresql.conf as an included shared object.
pg_upgrade needs access to any custom shared object files that are installed in
the "old" database. The upgrade image now includes the pgnodemx package for all
versions of postgres that are supported by crunchy repos.

There was also a permissions issue that caused the upgrade job to fail. pg_upgrade 
expects the old pgdata directory to have u=rwx (0700) permissions.
This change updates the permissions in the upgrade/start.sh to meet this
expectation and complete the upgrade.
[ch4854]

**Checklist:**

 
Tested centos7 and ubi8 in gke, tested ubi8 in gcp (ocp 3.11). Was able to follow upgrade example docs to complete a major upgrade (with test data)

**Type of Changes:**

 
**Other information**:
[ch9428]

### DIFF
--- a/bin/upgrade/start.sh
+++ b/bin/upgrade/start.sh
@@ -133,6 +133,9 @@ rm ${PGDATAOLD?}/postmaster.pid
 # changing to /tmp is necessary since pg_upgrade has to have write access
 cd /tmp
 
+# pg_upgrade expects the old pgdata directory to have u=rwx (0700) permissions
+chmod 0700 ${PGDATAOLD?}
+
 ${PGBINNEW?}/pg_upgrade
 rc=$?
 if (( $rc ==  0 )); then
--- a/bin/upgrade/start.sh
+++ b/bin/upgrade/start.sh
@@ -25,7 +25,8 @@
 # $NEW_VERSION (e.g. 9.6)
 #
 
-source /opt/cpm/bin/common_lib.sh
+CRUNCHY_DIR=${CRUNCHY_DIR:-'/opt/crunchy'}
+source "${CRUNCHY_DIR}/bin/common_lib.sh"
 enable_debugging
 
 function trap_sigterm() {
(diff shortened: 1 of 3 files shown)


bin/upgrade/start.sh
```diff
@@ -97,7 +98,7 @@ case $OLD_VERSION in
     ;;
 esac
 
-export PATH=/opt/cpm/bin:${PGBINNEW?}:$PATH
+export PATH=/opt/crunchy/bin:${PGBINNEW?}:$PATH
```
jmckulk: needs to user CRUNCHY_DIR

jkatz: ^

bin/upgrade/start.sh
```diff
@@ -132,6 +133,7 @@ rm ${PGDATAOLD?}/postmaster.pid
 # changing to /tmp is necessary since pg_upgrade has to have write access
 cd /tmp
 
+chmod 0700 ${PGDATAOLD?}
```
jkatz, line 137: Should leave a comment here as this is an important line.

jkatz (changes requested):

bin/upgrade/start.sh
```diff
@@ -97,7 +98,7 @@ case $OLD_VERSION in
     ;;
 esac
 
-export PATH=/opt/cpm/bin:${PGBINNEW?}:$PATH
+export "PATH=${CRUNCHY_DIR}/bin:${PGBINNEW?}:$PATH"
```
jkatz:
```suggestion
export PATH="${CRUNCHY_DIR}/bin:${PGBINNEW?}:${PATH}"
```

jkatz (changes requested):

jkatz (approved):

PR #1308, opened 2020-12-16
Move exporter function install to its own file

jkatz:
This allows for the external enablement of these functions, e.g.
from the Postgres Operator

Issue: [ch7270]

### DIFF
--- a/bin/postgres_common/exporter/install.sh
+++ b/bin/postgres_common/exporter/install.sh
@@ -1,0 +1,55 @@
+#!/bin/bash
+
+# Copyright 2019 - 2020 Crunchy Data Solutions, Inc.
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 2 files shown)


bin/postgres_common/exporter/install.sh
```diff
@@ -0,0 +1,55 @@
...
+
+echo_info "Using setup file '${function_file}' for pgMonitor"
+cp "${function_file}" "/tmp/setup_pg.sql"
+sed -i "s,/usr/bin/pgbackrest-info.sh,${CRUNCHY_DIR}/bin/postgres-ha/pgbackrest/pgbackrest_info.sh,g" "/tmp/setup_pg.sql"
+
+psql -U postgres --port="${PG_PRIMARY_PORT}" -d postgres \
+    < "/tmp/setup_pg.sql" > /tmp/pgmonitor-setup.stdout 2> /tmp/pgmonitor-setup.stderr
+
+psql -U postgres --port="${PG_PRIMARY_PORT}" -d postgres \
+    -c "CREATE EXTENSION IF NOT EXISTS pgnodemx WITH SCHEMA monitor;" > /tmp/pgmonitor-setup.stdout 2> /tmp/pgmonitor-setup.stderr
```
cbandy: Both of these calls overwrite the output files. Should one or both append, instead?

```sh
> /tmp/pgmonitor-setup.stdout 2> /tmp/pgmonitor-setup.stderr
```

cbandy (approved):

bin/postgres_common/exporter/install.sh
jkatz: Probably. That was outside the original scope of the file move, but I may as well fix them.

PR #1309, opened 2020-12-17
Modify JIT defaults

jkatz:
Given issues related to memory leaks with certain of the monitoring
queries being traced to JIT (CrunchyData/pgmonitor#182), JIT compilation
will be an opt-in feature.

Issue: [ch9928]

### DIFF
--- a/conf/postgres-ha/postgres-ha-bootstrap.yaml
+++ b/conf/postgres-ha/postgres-ha-bootstrap.yaml
@@ -10,6 +10,7 @@ bootstrap:
   dcs:
     postgresql:
       parameters:
+        jit: off
         unix_socket_directories: /tmp,/crunchyadm
         wal_level: logical
   post_bootstrap: /opt/crunchy/bin/postgres-ha/bootstrap/post-bootstrap.sh


andrewlecuyer (approved):

chobostar:
Modifying default could be not fair for whole database users. Some queries might need it.

I wonder is it possible to set `jit=off` to monitoring user? Currently postgres supports it, since setting in user's context:
```
postgres=# select name, context from pg_settings where name = 'jit';
-[ RECORD 1 ]-
name    | jit
context | user
```
like this:
```
alter user ccp_monitoring set jit=off;
```
Also it safer in case if someone changes default value to `on`.

cbandy:
@chobostar that's an excellent idea. PGO v5 disables `jit` by default, but one can override that in their PostgresCluster spec at any time. If you're willing to submit a PR, I believe the change belongs in [here](https://github.com/CrunchyData/postgres-operator/blob/330de87470d932dab5c2d54aca296e65577c3bf9/internal/pgmonitor/postgres.go#L92) probably near the existing `ALTER ROLE` statement.

chobostar:
@cbandy 
>If you're willing to submit a PR,If you're willing to submit a PR,

Yes, we have customers who use JIT, they were just lucky to have engineers who understand Postgres... I hope this approach will be more friendly to a wide audience of users
https://github.com/CrunchyData/postgres-operator/pull/3384

PR #1311, opened 2020-12-18
Compaction cleanup 1

jmckulk:
This PR is a collection of changes to tie together the Container Compaction series. 

The crunchy-admin container is updated to be based on crunchy-base instead of crunchy-pg-base. This is the final reference to crunchy-pg-base in the repository and references to image have now been removed. In the container compaction changes the default config and binary path in the images was updated from `/opt/cpm` to `/opt/crunchy`. This removes the last references to the old path from any accessory scripts in the repository. I have also removed old scripts that are no longer referenced in either the crunchy-containers or the postgres-operator repositories. 

The changes to remove the containers examples and associated documentation will be commit in a future pr.

**Checklist:**

 
**Type of Changes:**

 
**Other information**:
[ch9428]


build/upgrade/Dockerfile
```diff
@@ -10,37 +10,13 @@ ARG DFSET
...
-RUN rpm --import RPM-GPG-KEY-crunchydata*
-RUN if [ "$DFSET" = "rhel" ] ; then rpm --import CRUNCHY-GPG-KEY.public ; fi
-
-RUN if [ "$BASEOS" = "centos8" ] ; then \
-	${PACKAGER} -qy module disable postgresql ; \
-fi
-
-RUN if [ "$BASEOS" = "ubi8" ] ; then \
-	${PACKAGER} -qy module disable postgresql ; \
-fi
```
jkatz: We *might* still need these here? The upgrade container needs multiple PostgreSQL binaries.

build/pgbackrest/Dockerfile
```diff
@@ -1,31 +1,21 @@
...
 ARG BASEVER
 ARG PG_FULL
-ARG PG_MAJOR
 ARG PREFIX
 FROM ${PREFIX}/crunchy-base:${BASEOS}-${PG_FULL}-${BASEVER}
 
 # For RHEL8 all arguments used in main code has to be specified after FROM
-ARG DFSET
 ARG PACKAGER
 ARG PG_MAJOR
```
jmckulk: think this can be removed `ARG PG_MAJOR`

build/upgrade/Dockerfile
jmckulk: I did run into issues build the upgrade image without this. I can revert this change

jkatz (approved):

PR #1313, opened 2020-12-22
Update build order for PostGIS HA image

jkatz:
The PostGIS HA image now builds off of the base PostGIS image
instead of the base HA image, which should speed up the overall
build times.

### DIFF
--- a/build/postgres-gis-ha/Dockerfile
+++ b/build/postgres-gis-ha/Dockerfile
@@ -2,14 +2,24 @@ ARG BASEOS
 ARG BASEVER
 ARG PG_FULL
 ARG PREFIX
-FROM ${PREFIX}/crunchy-postgres-ha:${BASEOS}-${PG_FULL}-${BASEVER}
+ARG POSTGIS_VER
+FROM ${PREFIX}/crunchy-postgres-gis:${BASEOS}-${PG_FULL}-${POSTGIS_VER}-${BASEVER}
+# crunchy-postgres image runs as USER 26, switch back to install packages
+USER 0
 
+# ===== Early lines ordered for leveraging cache, reorder carefully =====
 # For RHEL8 all arguments used in main code has to be specified after FROM
 ARG BASEOS
+ARG PG_FULL
 ARG DFSET
 ARG PACKAGER
 ARG PG_MAJOR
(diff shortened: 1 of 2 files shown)


build/postgres-gis-ha/Dockerfile
```diff
@@ -18,92 +28,73 @@ LABEL name="postgres-gis-ha" \
...
+# Adjust ownership for the folders to be the "postgres" user and allow the group
+# permissions to match the user ones EXCEPT for the /tablespaces folder, which
+# will only have permissions on the user
+RUN chown -R postgres:postgres /crunchyadm /tablespaces && \
+	chmod -R g=u /crunchyadm /tablespaces
+
 # open up the postgres port
 EXPOSE 5432
 
-ADD bin/postgres-gis-ha /opt/crunchy/bin/postgres-ha
```
crunchyheath, line 106: I think we still need to add bin/postgres-gis-ha after we add the bin/postgres-ha.  Otherwise it doesn't create the postgis extensions, right?

jkatz: Oops you're right -- good catch.

build/postgres-gis-ha/Dockerfile
```diff
@@ -18,92 +28,74 @@ LABEL name="postgres-gis-ha" \
...
+# permissions to match the user ones EXCEPT for the /tablespaces folder, which
+# will only have permissions on the user
+RUN chown -R postgres:postgres /crunchyadm /tablespaces && \
+	chmod -R g=u /crunchyadm /tablespaces
+
 # open up the postgres port
 EXPOSE 5432
 
 ADD bin/postgres-gis-ha /opt/crunchy/bin/postgres-ha
+ADD bin/postgres-ha /opt/crunchy/bin/postgres-ha
```
cbandy: Both of these sources have a `sql/setup.sql`. As @crunchyheath was saying, `gis-ha` needs to be after `ha`.

jkatz: That leads me to many other questions/comments about the design, but in the interim I have made this adjustment.

crunchyheath (approved):
Looks good to me

PR #1314, opened 2020-12-24
Refactor language bases into base container

jkatz:
This moves the Perl/R dependencies into its own base to help
speed up the chain of building the various GIS-enabled containers.


crunchyheath (approved):
Looks good to me.  +1

PR #1315, opened 2020-12-26
Remove crunchy-admin container

jkatz:
This container was added provisionally in 4.2.0 to be used as
part of the authentication and administration scheme for the
Operator. However, this never was fully built out due to a variety
of reasons, and as such in 4.6.0 this is being abandoned.

A file found in `/crunchyadm/pgha_initialized` is still being used
as one of the Operator's readiness checks, and as such, this
directory is being kept for the time being.

Issue: [ch10007]
Related: CrunchyData/postgres-operator#2152


crunchyheath (approved):
To the best of my knowledge this seems okay.  +1

PR #1316, opened 2020-12-30
Update environmental variable references and code cleanup

jkatz:
The pgBackRest container now recognizes a more standard environmental
variable for working with pgBackRest and as such, the code needs to
be updated to recognize that.

Additionally, this presented an opportunity to clean up the code
a bit so it will be easier to amend in the future.

Issue: [ch9871]
Related: [https://github.com/CrunchyData/postgres-operator/pull/2158](https://github.com/CrunchyData/postgres-operator/pull/2158)


ValClarkson (approved):
lgtm

PR #1327, opened 2021-02-20
Use archive_command=false For New Cluster Restore

andrewlecuyer:
When a new cluster is being created using an existing repository, e.g. using command:

```bash
pgo create mycluster2 --restore-from=mycluster1
```

The `archive_command` is now set to `false` during the initial restore and Patroni bootstrap.

In support of this change, `archive_mode` is no longer set to `off` when performing a pgBackRest restore during the initialization of a new cluster, nor is pgBackRest stopped post-initialization.  These settings (which effectively disable archive pushing) prevent the `.history` file from being archived during the restore process as needed for the proper management of PG timelines, and setting `archive_mode=false` ensures both archives are not pushed while restoring and bootstrapping from an existing/running cluster (since two DB's should not be pushing to the same pgBackRest repository), while also ensuring archives (including history files) are then pushed when the proper `archive_command` is subsequently configured during the cluster initialization process.

**Checklist:**

 
**Type of Changes:**

 
`archive_mode` is set to `off` when performing a pgBackRest restore during the initialization of a new cluster, and pgBackRest is stopped post-initialization.

[ch10513]

`archive_mode` is no longer set to `off` when performing a pgBackRest restore during the initialization of a new cluster, nor is pgBackRest stopped post-initialization.  Additionally, when a new cluster is being created using using an existing repository the `archive_command` is now set to `false` during the initial restore. 

**Other information**:

N/A

### DIFF
--- a/bin/postgres-ha/bootstrap/pre-bootstrap.sh
+++ b/bin/postgres-ha/bootstrap/pre-bootstrap.sh
@@ -413,11 +413,8 @@ build_bootstrap_config_file() {
 
         if [[ "${bootstrap_cluster_source}" != "${bootstrap_cluster_target}" ]];
         then
-            echo_info "Disabling archive mode for bootstrap method ${PGHA_BOOTSTRAP_METHOD}"
-            "${CRUNCHY_DIR}/bin/yq" w -i "${bootstrap_file}" postgresql.parameters.archive_mode "off"
-        else
-            echo_info "Enabling archive mode for bootstrap method ${PGHA_BOOTSTRAP_METHOD}"
-            "${CRUNCHY_DIR}/bin/yq" w -i "${bootstrap_file}" postgresql.parameters.archive_mode "on"
+            echo_info "Disabling archiving for bootstrap method ${PGHA_BOOTSTRAP_METHOD}"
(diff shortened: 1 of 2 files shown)


bin/postgres-ha/bootstrap/pre-bootstrap.sh
```diff
@@ -414,10 +414,7 @@ build_bootstrap_config_file() {
         if [[ "${bootstrap_cluster_source}" != "${bootstrap_cluster_target}" ]];
         then
             echo_info "Disabling archive mode for bootstrap method ${PGHA_BOOTSTRAP_METHOD}"
```
jkatz: Let's update the message:

```suggestion
            echo_info "Disabling archiving for bootstrap method ${PGHA_BOOTSTRAP_METHOD}"
```

jkatz (approved):

bin/postgres-ha/bootstrap/pre-bootstrap.sh
andrewlecuyer: Thanks, good idea.  Message updated!

jkatz (approved):

PR #1332, opened 2021-03-08
Set system python for EL8 pgAdmin 4 containers to be python3

### LINKED ISSUE #2320: During pgadmin (centos8-12.6-4.6.1) backup/restore: python not found
**Describe the bug**
When I try to create a backup file or restore one, I got an error. python is not on the PATH in the image.

**To Reproduce**
1, Login to pgadmin
2, Right click on an object (schema, table, etc)
3, Backup or Restore
4, Select file
5, Click on Backup/Restore button

[Errno 2] No such file or directory: 'python': 'python'

**Please tell us about your environment:**

I use postgres operator 4.6.1 on OKD 4.6 with this image: crunchy-pgadmin4:centos8-12.6-4.6.1

**Additional context**
I found a workaround.
In the image there is no python symlink on the PATH. So I created a PATH environment variable and extend the existing PATH with /tmp.
New symlink in /tmp: /tmp/python->/usr/bin/python3
Now the backup and restore is working.

jkatz:
Python 2 is not installed in these, nor will it be, so this is a
safe measure, and it also allows for some specific pgAdmin 4
functionality to work.

Issue: [ch10718]
fixes CrunchyData/postgres-operator#2320


crunchyheath (commented):
+1 from me.   Surprised it wasn't already python3, but works for me

PR #1337, opened 2021-03-29
Adds the nss_wrapper for Arbitrary UID Support in OpenShift

andrewlecuyer:
Adds and enables the `nss_wrapper` for all containers using UID entrypoint scripts, specifically as needed for compatibility with the `restricted` SCC in OpenShift (and therefore the random UID's that are assigned to containers by OpenShift). This is done in accordance with RedHat guidance (see https://access.redhat.com/articles/4859371), specifically replacing the current solution which directly modifies the `/etc/passwd` and `/etc/group` files.

With this change, the `nss_wrapper` will now always be enabled when an entrypoint script (e.g. `uid_postgres.sh`, `uid_pgbackrest.sh`, etc.) is called. Additionally, the `/etc/passwd` and `/etc/group` files are no longer writeable, and entrypoint script naming has been maintained to ensure a seamless and transparent transition to the `nss_wrapper` solution (i.e. consumers of the containers are not required to make any changes to maintain current behavior and functionality).

**Checklist:**

 
**Type of Changes:**

 
Writeable `/etc/passwd` and `/etc/group` files are utilized to support arbitrary UID's in OpenShift.

[ch9887]

The `nss_wrapper` is utilized to support arbitrary UID's in OpenShift.

**Other information**:

Please note that this solution is intended to be compatible with all Docker and Kubernetes examples within the Crunchy Container Suite, as well as with all currently supported versions of the Crunchy PostgreSQL Operator.

### DIFF
--- a/bin/common/nss_wrapper.sh
+++ b/bin/common/nss_wrapper.sh
@@ -1,0 +1,71 @@
+#!/bin/bash
+
+# Copyright 2021 Crunchy Data Solutions, Inc.
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+CRUNCHY_DIR=${CRUNCHY_DIR:-'/opt/crunchy'}
+
+# The following sets up an nss_wrapper environment in accordance with OpenShift
+# guidance for supporting arbitrary user ID's
+
+# Define nss_wrapper directory and passwd & group files that will be utilized by nss_wrapper.  The
+# nss_wrapper_env.sh script (which also sets these vars) isn't sourced here since the nss_wrapper
+# has not yet been setup, and we therefore don't yet want the nss_wrapper vars in the environment.
+NSS_WRAPPER_DIR="/tmp/nss_wrapper/${NSS_WRAPPER_SUBDIR}"
+NSS_WRAPPER_PASSWD="${NSS_WRAPPER_DIR}/passwd"
+NSS_WRAPPER_GROUP="${NSS_WRAPPER_DIR}/group"
+
+# create the nss_wrapper directory
+mkdir -p "${NSS_WRAPPER_DIR}"
+
+# grab the current user ID and group ID
+USER_ID=$(id -u)
+export USER_ID
+GROUP_ID=$(id -g)
+export GROUP_ID
+
+# get copies of the passwd and group files
+[[ -f "${NSS_WRAPPER_PASSWD}" ]] || cp "/etc/passwd" "${NSS_WRAPPER_PASSWD}"
+[[ -f "${NSS_WRAPPER_GROUP}" ]] || cp "/etc/group" "${NSS_WRAPPER_GROUP}"
+
+# if the username is missing from the passwd file, then add it
+if [[ ! $(cat "${NSS_WRAPPER_PASSWD}") =~ ${CRUNCHY_NSS_USERNAME}:x:${USER_ID} ]]; then
+    echo "nss_wrapper: adding user"
+    passwd_tmp="${NSS_WRAPPER_DIR}/passwd_tmp"
+    cp "${NSS_WRAPPER_PASSWD}" "${passwd_tmp}"
+    sed -i "/${CRUNCHY_NSS_USERNAME}:x:/d" "${passwd_tmp}"
+    # needed for OCP 4.x because crio updates /etc/passwd with an entry for USER_ID
+    sed -i "/${USER_ID}:x:/d" "${passwd_tmp}"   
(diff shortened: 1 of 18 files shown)


bin/common/nss_wrapper.sh
```diff
@@ -0,0 +1,69 @@
...
+
+CRUNCHY_DIR=${CRUNCHY_DIR:-'/opt/crunchy'}
+
+# The following sets up an nss_wrapper environment in accordance with OpenShift
+# guidance for supporting arbitrary user ID's
+
+# define nss_wrapper directory and passwd & group files that will be utilized by nss_wrapper
+NSS_WRAPPER_DIR="/tmp/nss_wrapper/${NSS_WRAPPER_SUBDIR}"
+NSS_WRAPPER_PASSWD="${NSS_WRAPPER_DIR}/passwd"
+NSS_WRAPPER_GROUP="${NSS_WRAPPER_DIR}/group"
```
jkatz: @andrewlecuyer shouldn't we just call `source nss_wrapper_env.sh` here?

bin/pgbackrest-repo/pgbackrest-repo.sh
```diff
@@ -67,10 +81,8 @@ then
 	
 fi
 
-mkdir -p ~/.ssh/
-cp $CONFIG/config ~/.ssh/
 cp $CONFIG/id_ed25519 /tmp
-chmod 400 /tmp/id_ed25519 ~/.ssh/config
```
jkatz, line 73: This _might_ let us clear one of the emptyDirs, so noting that.

build/postgres-gis-ha/Dockerfile
```diff
@@ -95,6 +95,9 @@ RUN chmod +x /opt/crunchy/bin/yq
 # volume permissions are applied when building the image
 VOLUME ["/pgdata", "/pgwal", "/pgconf", "/backrestrepo", "/sshd"]
 
+# Defines a unique directory name that will be utilized by the nss_wrapper in the UID script
+ENV NSS_WRAPPER_SUBDIR="postgis-ha"
+
```
jkatz: Are we unable to leverage this from the `postgres-gis` container?

build/postgres-ha/Dockerfile
```diff
@@ -94,6 +94,9 @@ RUN chmod +x /opt/crunchy/bin/yq
 # volume permissions are applied when building the image
 VOLUME ["/pgdata", "/pgwal", "/pgconf", "/backrestrepo", "/sshd"]
 
+# Defines a unique directory name that will be utilized by the nss_wrapper in the UID script
+ENV NSS_WRAPPER_SUBDIR="postgres-ha"
+
```
jkatz: Are we unable to leverage this from the `postgres` container?

build/postgres-gis-ha/Dockerfile
andrewlecuyer: In general that should be fine, since uniqueness is really to avoid conflicts when two containers are running in the same Pod, which obviously wouldn't happen with `postgres-gis` & `postgres-gis-ha`.

However, I felt that establishing unique directories per-container keeps us safe regardless of how these containers are run, so I simply decided to add a unique value for this container as well.

build/postgres-ha/Dockerfile
andrewlecuyer: See comment for `postgres-gis-ha`.

bin/pgbackrest-repo/pgbackrest-repo.sh
andrewlecuyer: I did remove the `pgbackrest-ssh` volume, which was mounted to `/var/lib/pgsql/.ssh`.

bin/common/nss_wrapper.sh
jkatz: OK. May be good to add a comment to that effect so someone does not attempt to optimize it.

build/postgres-gis-ha/Dockerfile
jkatz: OK.

jkatz (approved):
Provisionally approved, though would like to see a comment added where indicated.

build/postgres-gis-ha/Dockerfile
jkatz: Actually I take this back -- Looking at https://github.com/CrunchyData/postgres-operator/pull/2356/files#diff-9438712811604f4e58616aad9619188a102937147f58ec5745bb9903f55d0284R155 this seems like it will break things.

jkatz (changes requested):
I think those ENV VARs need tobe changed, looking at the config templates in the Operator.

build/postgres-gis-ha/Dockerfile
andrewlecuyer: Thansk, great catch - I will adjust to use the same name across _all_ PG containers.

bin/common/nss_wrapper.sh
andrewlecuyer: Comment added.

build/postgres-gis-ha/Dockerfile
andrewlecuyer: The `crunchy-postgres-ha`, `crunchy-postgres-gis` and `crunchy-postgres-gis-ha` containers now all simply inherit the `NSS_WRAPPER_SUBDIR` variable defined in the `postgres` Dockerfile.

jkatz (approved):

PR #1342, opened 2021-04-13
Support ubi8-minimal

jkatz:
Upstream recommends the usage of ubi8-minimal images as a base
instead of the full ubi8 images, so this patch allows us to do
so.


crunchyheath (approved):
This looks good to me

PR #1351, opened 2021-04-23
Ensure proper ownership of NSS wrapper directory

jkatz:
This case can occur on vanilla Kubernetes clusters when there
are multiple sidecars in a Pod.


bin/common/nss_wrapper.sh
```diff
@@ -21,6 +21,9 @@ CRUNCHY_DIR=${CRUNCHY_DIR:-'/opt/crunchy'}
 # Define nss_wrapper directory and passwd & group files that will be utilized by nss_wrapper.  The
 # nss_wrapper_env.sh script (which also sets these vars) isn't sourced here since the nss_wrapper
 # has not yet been setup, and we therefore don't yet want the nss_wrapper vars in the environment.
+mkdir /tmp/nss_wrapper
+chmod g+rwx /tmp/nss_wrapper
```
cbandy, line 25: ⛏️ `install` can do this in one go, though it does have to be the entire permissions:

```suggestion
install -m 0775 -d /tmp/nss_wrapper
```

or

```sh
install -m a+rx,ug+w -d /tmp/nss_wrapper
```

andrewlecuyer (approved):

PR #1355, opened 2021-04-28
Ensure tar present in pgBackRest UBI 8 containers

jkatz:
tar is not present in the pgBackRest container for UBI 8,
but is now available.

This also ensures that bzip2 + lz4 are available, should one
select those modes.

Issue: [ch11325]


crunchyheath (approved):
+1 from my perspective

PR #1356, opened 2021-04-28
Ensure tar present in Postgres UBI 8 containers

jkatz:
Like 28068ca9, tar is now present in the Postgres container for UBI 8.

This also ensures that bzip2 + lz4 are available, should one select
those modes.


crunchyheath (approved):
Looks good to me

PR #1397, opened 2021-10-28
Fix PG Installs in UBI 8 Upgrade Container

andrewlecuyer:
Provides a solution for installing all required PG packages in the UBI 8 `crunchy-upgrade` container, specifically to address a limitation in `microdnf` in which it is not possible to use regular expressions in package names when installing a package using `microdnf install`.

This solution maintains existing build behavior in which the versions of PG installed in the upgrade container are determined by the `.repo` files present (and therefore included in the container) when building.

_**Note:** This change will need to be back-patched to the `REL_4_7` branch._

[sc-12978]

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -243,6 +243,32 @@ endif
 
 pgbackrest-pgimg-docker: pgbackrest-pgimg-build
 
+# ----- Special case image (upgrade) -----
+
+# Special case args: UPGRADE_PG_VERSIONS (defines all versions of PG that will be installed) 
+upgrade-img-build: ccbase-image $(CCPROOT)/build/upgrade/Dockerfile
+	$(IMGCMDSTEM) \
+		-f $(CCPROOT)/build/upgrade/Dockerfile \
+		-t $(CCP_IMAGE_PREFIX)/crunchy-upgrade:$(CCP_IMAGE_TAG) \
+		--build-arg BASEOS=$(CCP_BASEOS) \
+		--build-arg BASEVER=$(CCP_VERSION) \
+		--build-arg PG_FULL=$(CCP_PG_FULLVERSION) \
+		--build-arg PG_MAJOR=$(CCP_PGVERSION) \
+		--build-arg PREFIX=$(CCP_IMAGE_PREFIX) \
+		--build-arg DFSET=$(DFSET) \
+		--build-arg PACKAGER=$(PACKAGER) \
+		--build-arg UPGRADE_PG_VERSIONS="$(shell find $(CCPROOT)/conf -type f -name "crunchypg*.repo" | \
+			grep -o [1-9][0-9])" \
+		$(CCPROOT)
+
+upgrade-img-buildah: upgrade-img-build ;
+# only push to docker daemon if variable IMG_PUSH_TO_DOCKER_DAEMON is set to "true"
+ifeq ("$(IMG_PUSH_TO_DOCKER_DAEMON)", "true")
+	sudo --preserve-env buildah push $(CCP_IMAGE_PREFIX)/crunchy-upgrade:$(CCP_IMAGE_TAG) docker-daemon:$(CCP_IMAGE_PREFIX)/crunchy-upgrade:$(CCP_IMAGE_TAG)
+endif
+
+upgrade-img-docker: upgrade-img-build
+
 # ----- Extra images -----
 %-img-build: ccbase-image $(CCPROOT)/build/%/Dockerfile
 	$(IMGCMDSTEM) \
--- a/build/upgrade/Dockerfile
+++ b/build/upgrade/Dockerfile
@@ -9,6 +9,7 @@ ARG BASEOS
 ARG DFSET
 ARG PACKAGER
 ARG PG_MAJOR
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -243,6 +243,32 @@ endif
...
+		-t $(CCP_IMAGE_PREFIX)/crunchy-upgrade:$(CCP_IMAGE_TAG) \
+		--build-arg BASEOS=$(CCP_BASEOS) \
+		--build-arg BASEVER=$(CCP_VERSION) \
+		--build-arg PG_FULL=$(CCP_PG_FULLVERSION) \
+		--build-arg PG_MAJOR=$(CCP_PGVERSION) \
+		--build-arg PREFIX=$(CCP_IMAGE_PREFIX) \
+		--build-arg DFSET=$(DFSET) \
+		--build-arg PACKAGER=$(PACKAGER) \
+		--build-arg UPGRADE_PG_VERSIONS="$(shell find $(CCPROOT)/conf -type f -name "crunchypg*.repo" | \
+			grep -o [1-9][0-9] | tr '\n' ' ')" \
```
cbandy: ⛏️ I think `$(shell)` already replaces newlines with space. https://www.gnu.org/software/make/manual/html_node/Shell-Function.html#Shell-Function

```suggestion
			grep -o [1-9][0-9])" \
```

build/upgrade/Dockerfile
```diff
@@ -49,28 +50,30 @@ fi
...
-		&& ${PACKAGER} -y clean all ; \
-fi
+RUN for pg_version in $UPGRADE_PG_VERSIONS; do \
+        if [ "$BASEOS" = "ubi8" ] ; then \
+            ${PACKAGER} -y install --nodocs \
+                --disablerepo=* \
+                --enablerepo=crunchypg* \
+                postgresql$pg_version \
+                postgresql$pg_version-contrib \
+                postgresql$pg_version-server \
```
cbandy: ⛏️ This is a bit tricky for me to read. ShellCheck has convinced me to [use brackets around every variable](https://github.com/koalaman/shellcheck/wiki/SC2250) so the reader (me, today) doesn't have to remember what is or isn't part of a var name.

Do you mind changing all the `$pg_version` to `${pg_version}`?

build/upgrade/Dockerfile
```diff
@@ -49,28 +50,30 @@ fi
...
+            ${PACKAGER} -y install --nodocs \
+                --setopt=skip_missing_names_on_install=False \
+                --disablerepo=* \
+                --enablerepo=crunchypg* \
+                postgresql$pg_version \
+                postgresql$pg_version-contrib \
+                postgresql$pg_version-server \
+                pgaudit$pg_version \
+                pgnodemx$pg_version \
+            && ${PACKAGER} -y clean all ; \
```
cbandy: 🤔 Does this `clean all` inside the loop mean that the package manager has to download metadata every time? If so, perhaps it can go at the end instead?

Like, (pseudo)

```
for …; do
  …
done && ${PACKAGER} -y clean all
```

Makefile
andrewlecuyer: Yup, that works.  Updated.

build/upgrade/Dockerfile
andrewlecuyer: Brackets added.

build/upgrade/Dockerfile
andrewlecuyer: Makes sense, moved this as suggested.

Makefile
```diff
@@ -243,6 +243,32 @@ endif
...
+		--build-arg DFSET=$(DFSET) \
+		--build-arg PACKAGER=$(PACKAGER) \
+		--build-arg UPGRADE_PG_VERSIONS="$(shell find $(CCPROOT)/conf -type f -name "crunchypg*.repo" | \
+			grep -o [1-9][0-9])" \
+		$(CCPROOT)
+
+upgrade-img-buildah: upgrade-img-build ;
+# only push to docker daemon if variable IMG_PUSH_TO_DOCKER_DAEMON is set to "true"
+ifeq ("$(IMG_PUSH_TO_DOCKER_DAEMON)", "true")
+	sudo --preserve-env buildah push $(CCP_IMAGE_PREFIX)/crunchy-upgrade:$(CCP_IMAGE_TAG) docker-daemon:$(CCP_IMAGE_PREFIX)/crunchy-pgbackrest:$(CCP_IMAGE_TAG)
```
tjmoore4: :wrench: Looks like a copy/paste error.
```suggestion
	sudo --preserve-env buildah push $(CCP_IMAGE_PREFIX)/crunchy-upgrade:$(CCP_IMAGE_TAG) docker-daemon:$(CCP_IMAGE_PREFIX)/crunchy-upgrade:$(CCP_IMAGE_TAG)
```

tjmoore4 (commented):
:+1: Looks good overall, one issue.

Makefile
andrewlecuyer: Thanks, the image name has been update as suggested.

tjmoore4 (approved):
:+1: With update, looks good to me.

PR #1406, opened 2021-12-03
Add krb5-workstation Package

andrewlecuyer:
Adds the `krb5-workstation` package to the `crunchy-postgres` and `crunchy-upgrade` containers as required for Kerberos authentication.

[sc-13235]

### DIFF
--- a/build/postgres/Dockerfile
+++ b/build/postgres/Dockerfile
@@ -59,6 +59,7 @@ RUN if [ "$DFSET" = "centos" ] ; then \
         	unzip \
         	bzip2 \
         	lz4 \
+			krb5-workstation \
         && ${PACKAGER} -y clean all ; \
 else \
         ${PACKAGER} -y install --nodocs \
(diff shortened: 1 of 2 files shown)


build/upgrade/Dockerfile
```diff
@@ -45,6 +46,7 @@ else \
 		"pgaudit${PG_MAJOR//.}*" \
 		"pgnodemx${PG_MAJOR//.}" \
 		unzip \
+		krb5-workstation \
```
cbandy: ⛏️ Does it matter what order the packages are listed? If not, I'd prefer they stay alphabetical.

build/postgres/Dockerfile
```diff
@@ -87,6 +88,7 @@ else \
 		tar \
 		bzip2 \
 		lz4 \
+		krb5-workstation \
```
cbandy, line 91: 🤔 This list doesn't seem to be in any particular order! 🤷🏻

cbandy (approved):

build/upgrade/Dockerfile
andrewlecuyer: Package has been properly alphabetized in this Dockerfile.

build/postgres/Dockerfile
andrewlecuyer: Random order has been maintained for now.

PR #1412, opened 2022-02-10
Pin pgAdmin Version in Makefile

andrewlecuyer:
Pins the pgAdmin version in the Makefile. 

Also updates the pgAdmin Make targets and Dockerfile to accommodate an argument for providing a specific pgAdmin version.

This same change will need to be backpatched to v4.

[sc-13733]


crunchyheath (approved):
This looks good to me.  +1

cbandy (approved):

PR #1488, opened 2022-11-15
Updated pg upgrade Dockerfile to import all pg extensions.

dsessler7:
**Checklist:**

 
**Type of Changes:**

 
Currently, the `crunchy-upgrade` image does not have all of the pg extensions installed. Therefore, when a pg cluster that is using any of the extensions that are not installed in the `crunchy-upgrade` image is upgraded, the upgrade will fail. 

[sc-16369]

All pg extensions are now installed when a `crunchy-upgrade` image is built. Therefore, any pg clusters using any extensions can be successfully upgraded.

**Other information**:

### DIFF
--- a/build/upgrade/Dockerfile
+++ b/build/upgrade/Dockerfile
@@ -31,9 +31,17 @@ RUN if [ "$BASEOS" = "ubi8" ] ; then \
 		"postgresql${PG_MAJOR//.}" \
 		"postgresql${PG_MAJOR//.}-contrib" \
 		"postgresql${PG_MAJOR//.}-server" \
+		postgresql${PG_MAJOR//.}-plpython* \
 		"pgaudit${PG_MAJOR//.}*" \
+		pgaudit${PG_MAJOR//.}_set_user \
 		"pgnodemx${PG_MAJOR//.}" \
 		"orafce_${PG_MAJOR//.}" \
+		pg_partman_${PG_MAJOR//.} \
+		pg_cron_${PG_MAJOR//.} \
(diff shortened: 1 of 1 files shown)


build/upgrade/Dockerfile
```diff
@@ -45,9 +52,16 @@ else \
 		"postgresql${PG_MAJOR//.}" \
 		"postgresql${PG_MAJOR//.}-contrib" \
 		"postgresql${PG_MAJOR//.}-server" \
+		postgresql${PG_MAJOR//.}-plpython* \
 		"pgaudit${PG_MAJOR//.}*" \
```
tjmoore4, line 57: :question: Should `pgaudit${PG_MAJOR//.}_set_user` be set here as well?

tjmoore4 (commented):
Looks good overall, one question. Assuming all the testing passed as expected, this should be just about ready to go.

build/upgrade/Dockerfile
dsessler7: Probably... I think I thought that the asterisk in `"pgaudit${PG_MAJOR//.}*" \` was acting as a wildcard and that the `_set_user` would be grabbed as well? But I'm probably just confused on what that `*` is doing

tjmoore4: Ah, I think I missed the `*`. Yes, I believe that asterisk should work, but I haven't checked it locally. I'd be curious to know exactly what that installs.

tjmoore4 (approved):
With that question answered, I believe this should be good to go (assuming the `pgaudit_set_user` package is installed as expected).

PR #1491, opened 2022-12-02
PG 15 updates

jmckulk:
With pg15, we will need to import a new GPG key to build our images. Copy the new key into our base image.
**Checklist:**

 
 - [ ] Have you added an explanation of what your changes do and why you'd like them to be included?

**Type of Changes:**

 
**Other information**:
[sc-16072]
[sc-16925]

### DIFF
--- a/build/postgres/Dockerfile
+++ b/build/postgres/Dockerfile
@@ -49,7 +49,7 @@ RUN if [ "$DFSET" = "centos" ] ; then \
         	python3-psutil \
         	python3-psycopg2 \
         	rsync \
-        	$( printf '11\n'${PG_MAJOR} | sort -VC && echo timescaledb_${PG_MAJOR} ) \
+        	$( printf '10\n'${PG_MAJOR}'\n15' | sort -uVC && echo timescaledb_${PG_MAJOR} ) \
         	wal2json_${PG_MAJOR//.} \
         	file \
         	gettext \
@@ -82,7 +82,7 @@ else \
 		python3-psutil \
 		python3-psycopg2 \
 		rsync \
-		$( printf '11\n'${PG_MAJOR} | sort -VC && echo timescaledb_${PG_MAJOR} ) \
+		$( printf '10\n'${PG_MAJOR}'\n15' | sort -uVC && echo timescaledb_${PG_MAJOR} ) \
 		wal2json_${PG_MAJOR//.} \
 		file \
 		unzip \
--- a/build/base/Dockerfile
+++ b/build/base/Dockerfile
@@ -81,7 +81,7 @@ ADD conf/*KEY* /
 # Add any available Crunchy PG repo files
 ADD conf/crunchypg${PG_LBL}.repo /etc/yum.repos.d/
 # Import both keys to support all required repos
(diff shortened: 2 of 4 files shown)


build/postgres/Dockerfile
```diff
@@ -82,7 +81,6 @@ else \
 		python3-psutil \
 		python3-psycopg2 \
 		rsync \
-		$( printf '11\n'${PG_MAJOR} | sort -VC && echo timescaledb_${PG_MAJOR} ) \
```
cbandy, line 85: 🤔 I know these aren't easy to comprehend. It's unfortunate we can't comment around here, either. This was doing "11 <= PG_MAJOR."

📝 We could update it to "11 <= PG_MAJOR <= 14" with 
```sh
printf '11\n'${PG_MAJOR}'\n14' | sort -VC
```

or to "10 < PG_MAJOR < 15" with
```sh
printf '10\n'${PG_MAJOR}'\n14' | sort -uVC
```

I prefer the former.

build/postgres/Dockerfile
```diff
@@ -98,6 +96,13 @@ else \
 	&& ${PACKAGER} -y clean all --enablerepo="epel" ; \
 fi
 
+RUN if [ "${PG_MAJOR}" -lt "15" ]; then \
+	${PACKAGER} -y install --nodocs \
+		--disablerepo=* \
+		--enablerepo=crunchypg* \
+		$( printf '11\n'${PG_MAJOR} | sort -VC && echo timescaledb_${PG_MAJOR} ) ; \
```
cbandy: 🤔 This doesn't seem quite right. We don't build PG 10 now, but `PG_MAJOR=10` would run this line and the command would be ... `$PACKAGER install # nothing?`

If we're adding `RUN` commands and `if` statements, it's probably best to do all the logic there.

```sh
if [ "${PG_MAJOR}" -ge 11 ] && [ "${PG_MAJOR}" -lt 15 ]; then
```

cbandy: Ah, and now I remember why we weren't using these before. We used to build for 9.5, 9.6, etc. `sort -V` does non-integer (version) comparison.

build/postgres/Dockerfile
jmckulk: Oh wow, I didn't realize this is what we were doing. A comment here would have been helpful. I can update to use this logic instead of a new IF statement

build/postgres/Dockerfile
jmckulk: Again, I didn't realize what the `sort -V` was doing. If I move the logic back into the original `sort`, this is a non-issue?

build/postgres/Dockerfile
cbandy: 🙇🏻 ⛏️ I think I'm finding I more than "prefer" using `11` and `14`. I read it as "for 11 through 14, timescale." I'm not sure how to read `10` and `15`... "for everything but 10 ... or 15 ..."

It's a nitpick at this point. The logic is sound.

cbandy (approved):

